Issue #18, opened 2021-11-26
Add containerized build support


daniel-hutao:
## Description

We should move the compile/build process to a container to ensure that everyone can get consistent products. Local compile/build often introduces problems due to inconsistent environment, such as inconsistent golang versions.

## Describe the Proposal

1. Adding golang:1.17.x as a builder container.
2. Updating Dockerfile & Makefile.
3. ...

PR #20, opened 2021-11-26
feat: Add containerized build support

daniel-hutao:
# Summary

## Key Points

## Description

use `make docker-build` to build, bin files is located at **build/output/**

截屏2021-11-26 上午10 15 49

the benefit with cache dips layer

截屏2021-11-26 上午11 42 35

### Related Issues

fix: #18


daniel-hutao:
@basicthinker @IronCore864 Please take a look when free.

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM golang:1.17 as builder
+
+ENV GOPROXY=https://goproxy.cn
+
+WORKDIR /devstream
```
IronCore864, line 5: After second thought, I think there might be a few issues with building in docker:

## 1. Network Environment

Building inside docker (instead of on their machines directly) for users behind a corporate proxy might cause a `certificate signed by unknown authority` issue, depending on their corporate proxy setup. This wouldn't be a big issue if they built it locally because the chances are, they've already solved it locally before (like when they used `curl`, for example).

## 2. Cross Platform

In essence, the build inside the docker is for Debian/Ubuntu (depending on the base image); but after the build, the binary is copied to the host of the user and will be executed there, and the users' host might differ; maybe it's Mac, maybe it's Windows.

Theoretically, we can write some shell script to detect if the user needs proxy, and what's the user's OS, then call different build commands like:

- `make build-in-docker-without-cn-proxy-debian`
- `make build-in-docker-without-cn-proxy-darwin`
- `make build-in-docker-without-cn-proxy-windows`

- `make build-in-docker-with-cn-proxy-debian`
- `make build-in-docker-with-cn-proxy-darwin`
- `make build-in-docker-with-cn-proxy-windows`

This might be a bit redundant, though.

I'm not completely sure if this is the simplest way to build a CLI tool.

@lfbdev

IronCore864 (changes requested):
See comments inline, especially about the "build in docker" part. We need a bit of discussion there.

Makefile
```diff
@@ -1,6 +1,17 @@
...
+
 build:
 	go get ./...
 	go build -buildmode=plugin -trimpath -gcflags="all=-N -l" -o plugins/githubactions_0.0.1.so ./cmd/githubactions/
 	go build -buildmode=plugin -trimpath -gcflags="all=-N -l" -o plugins/argocd_0.0.1.so ./cmd/argocd/
 	go build -buildmode=plugin -trimpath -gcflags="all=-N -l" -o plugins/argocdapp_0.0.1.so ./cmd/argocdapp/
 	go build -trimpath -gcflags="all=-N -l" -o dsm ./cmd/devstream/
+
+# dsm and plugins will be located at build/output/
+docker-build:
```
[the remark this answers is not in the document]
IronCore864: Reminder: the README.md also needs to be updated accordingly, but please hold on till we resolve the "build in docker" discussion first.

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM golang:1.17 as builder
+
+ENV GOPROXY=https://goproxy.cn
```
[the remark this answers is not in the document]
daniel-hutao: OK, I'll add a parameter for this later.

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+WORKDIR /devstream
+
+# cache deps before building so that source changes don't invalidate our downloaded layer
+COPY go.mod go.mod
+
+COPY go.sum go.sum
```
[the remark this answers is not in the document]
daniel-hutao: You're right, it can be Integrate to 1 line. I wrote it 2 lines because it's the default style by kubebuilder.

Makefile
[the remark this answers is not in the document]
daniel-hutao: Can we rename `docker-build` to `build` and remove the `build` part? And I'll add the parameter to choose GOPROXY at Makefile.

build.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+GOOS=linux GOARCH=amd64 go build -buildmode=plugin -trimpath -gcflags="all=-N -l" -o output/plugins/githubactions_0.0.1.so ./cmd/githubactions/
+GOOS=linux GOARCH=amd64 go build -buildmode=plugin -trimpath -gcflags="all=-N -l" -o output/plugins/argocd_0.0.1.so ./cmd/argocd/
+GOOS=linux GOARCH=amd64 go build -buildmode=plugin -trimpath -gcflags="all=-N -l" -o output/plugins/argocdapp_0.0.1.so ./cmd/argocdapp/
+GOOS=linux GOARCH=amd64 go build -trimpath -gcflags="all=-N -l" -o output/dsm ./cmd/devstream/
```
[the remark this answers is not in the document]
daniel-hutao: Good suggestion

Dockerfile
daniel-hutao: Yes, I also have the idea to support cross-platform compile, but support the linux/amd64 first, and I'll add mac support recently

Makefile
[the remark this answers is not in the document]
daniel-hutao: Or we can name it with build-docker-linux-amd64 first, then gradually add support for other platforms.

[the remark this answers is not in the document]
daniel-hutao: Sorry, I can get the GOOS & GOARCH from user's os env, then set it to the build.sh

daniel-hutao:
@IronCore864 I've updated the code with suggestions above. Only one problem left is the image **golang:1.17** is not support Darwin os. 

In the near future, I will continue to look for a suitable environment base image, or build a base image that supports multi-platform compile by myself, and at that time, I'll turn back to  Complete the multi-platform compile feature.

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM golang:1.17 as builder
+
+ARG GOPROXY=https://goproxy.io
+
+ENV GOPROXY=${GOPROXY}
```
IronCore864, line 5: The name of the ARG and the ENV is the same; it might not be the clearest way.

How about `ARG GOPROXY_ARG`, and `ENV GOPROXY`?

Makefile
```diff
@@ -1,6 +1,21 @@
+GOPROXY ?= https://goproxy.io
```
IronCore864, line 1: On both the `Dockerfile` and the `Makefile`: I think it's better to make "no proxy" as the default behavior (I.E., the default for global users, instead of for users who are behind the GFW without a VPN). The reason behind this is, we want to make it convenient for _the majority_ of the developers.

For developers behind the GFW without a VPN, they probably already have GOPROXY ENV var set in their local environment, so they can just run `make build-with-docker` without any change or extra parameter.

Makefile
```diff
@@ -1,6 +1,21 @@
...
+GOOS ?= linux
+GOARCH ?= amd64
```
IronCore864, line 3: Regarding the default behavior for OS and arch, my bet is, we should set the default as darwin/arm64 (M1 Mac users) instead of Linux/amd64.

Makefile
```diff
@@ -1,6 +1,21 @@
...
 	go build -buildmode=plugin -trimpath -gcflags="all=-N -l" -o plugins/argocdapp_0.0.1.so ./cmd/argocdapp/
 	go build -trimpath -gcflags="all=-N -l" -o dsm ./cmd/devstream/
+
+# dsm and plugins will be located at build/output/
+build-with-docker:
+	rm -rf build/ && mkdir build
+	# update the deps layer if needed
+	docker build --build-arg GOPROXY=${GOPROXY} -t devstream_build:v0.1 .
+	cp -r go.mod go.sum cmd internal build.sh build/
+	docker run -v ${ROOT_PATH}build:/devstream -e GOOS=${GOOS} -e GOARCH=${GOARCH} --rm devstream_build:v0.1
```
IronCore864, line 21: Line 19, 21: let's make it `0.0.1`, which is the same as the plugins' versions. I think it's customary that the docker image tag doesn't have the leading "v" in front of the version number, while the git tag has it.

On the docker image tag, maybe we can consider `mericodev/stream-builder` given that we already have `[mericodev/lake-builder](https://hub.docker.com/r/mericodev/lake-builder)`

IronCore864 (commented):
Please hold on for a while before changing anything because I'm running more tests locally with this PR (mainly w/ or w/o proxy, VPN, etc.)

@lfbdev please also have a look and comment.

IronCore864:
I'm having this error locally (with GOPROXY, without VPN, without changing anything from this PR:

```bash
tiexin@MBA ~/work/hutao/stream $ make build-with-docker
rm -rf build/ && mkdir build
# update the deps layer if needed
docker build --build-arg GOPROXY=https://goproxy.io -t devstream_build:v0.1 .
[+] Building 103.8s (9/9) FINISHED                                                                                                                                                                                        
 => [internal] load build definition from Dockerfile                                                                                                                                                                 0.0s
 => => transferring dockerfile: 311B                                                                                                                                                                                 0.0s
 => [internal] load .dockerignore                                                                                                                                                                                    0.0s
 => => transferring context: 2B                                                                                                                                                                                      0.0s
 => [internal] load metadata for docker.io/library/golang:1.17                                                                                                                                                       0.0s
 => [1/4] FROM docker.io/library/golang:1.17                                                                                                                                                                         0.0s
 => [internal] load build context                                                                                                                                                                                    0.0s
 => => transferring context: 55B                                                                                                                                                                                     0.0s
 => CACHED [2/4] WORKDIR /devstream                                                                                                                                                                                  0.0s
 => CACHED [3/4] COPY go.mod go.sum /devstream/                                                                                                                                                                      0.0s
 => [4/4] RUN go mod download                                                                                                                                                                                       90.6s
 => exporting to image                                                                                                                                                                                              13.2s
 => => exporting layers                                                                                                                                                                                             13.1s
 => => writing image sha256:6a06db6b4aa8e94ade6e7a6b7dae008dfabb4e68c8902c04fb2f214583243ecb                                                                                                                         0.0s
 => => naming to docker.io/library/devstream_build:v0.1                                                                                                                                                              0.0s

Use 'docker scan' to run Snyk tests against images to find vulnerabilities and learn how to fix them
cp -r go.mod go.sum cmd internal build.sh build/
docker run -v /Users/tiexin/work/hutao/stream/build:/devstream -e GOOS=linux -e GOARCH=amd64 --rm devstream_build:v0.1
# github.com/merico-dev/stream/cmd/githubactions
loadinternal: cannot find runtime/cgo
/usr/local/go/pkg/tool/linux_arm64/link: running gcc failed: exit status 1
gcc: error: unrecognized command-line option '-m64'

make: *** [build-with-docker] Error 2
```

@daniel-hutao please have a look.

IronCore864:
@IronCore864 a reminder to myself: the README needs to be updated with more detailed build steps; no need to do in this PR; will add later separately.

Dockerfile
daniel-hutao: Ok, xx_ARG is clear

Makefile
daniel-hutao: It's not very accurate, the GOPROXY ENV is used inside the container, the env set in their local environment is not work.

If we make "no proxy" as the default behavior, many developers will get failed, we can consider it more.

Makefile
daniel-hutao: I don’t know which user scenario is more for mac and linux. Maybe there are more macs for writing code and more linux for running. We can discuss it more.

Makefile
daniel-hutao: stream-builder is better, but maybe it will be better to do it later

Dockerfile
daniel-hutao: > I'm having this error locally (with GOPROXY, without VPN, without changing anything from this PR, M1 Mac):
> 
> ```shell
> tiexin@MBA ~/work/hutao/stream $ make build-with-docker
> rm -rf build/ && mkdir build
> # update the deps layer if needed
> docker build --build-arg GOPROXY=https://goproxy.io -t devstream_build:v0.1 .
> [+] Building 103.8s (9/9) FINISHED                                                                                                                                                                                        
>  => [internal] load build definition from Dockerfile                                                                                                                                                                 0.0s
>  => => transferring dockerfile: 311B                                                                                                                                                                                 0.0s
>  => [internal] load .dockerignore                                                                                                                                                                                    0.0s
>  => => transferring context: 2B                                                                                                                                                                                      0.0s
>  => [internal] load metadata for docker.io/library/golang:1.17                                                                                                                                                       0.0s
>  => [1/4] FROM docker.io/library/golang:1.17                                                                                                                                                                         0.0s
>  => [internal] load build context                                                                                                                                                                                    0.0s
>  => => transferring context: 55B                                                                                                                                                                                     0.0s
>  => CACHED [2/4] WORKDIR /devstream                                                                                                                                                                                  0.0s
>  => CACHED [3/4] COPY go.mod go.sum /devstream/                                                                                                                                                                      0.0s
>  => [4/4] RUN go mod download                                                                                                                                                                                       90.6s
>  => exporting to image                                                                                                                                                                                              13.2s
>  => => exporting layers                                                                                                                                                                                             13.1s
>  => => writing image sha256:6a06db6b4aa8e94ade6e7a6b7dae008dfabb4e68c8902c04fb2f214583243ecb                                                                                                                         0.0s
>  => => naming to docker.io/library/devstream_build:v0.1                                                                                                                                                              0.0s
> 
> Use 'docker scan' to run Snyk tests against images to find vulnerabilities and learn how to fix them
> cp -r go.mod go.sum cmd internal build.sh build/
> docker run -v /Users/tiexin/work/hutao/stream/build:/devstream -e GOOS=linux -e GOARCH=amd64 --rm devstream_build:v0.1
> # github.com/merico-dev/stream/cmd/githubactions
> loadinternal: cannot find runtime/cgo
> /usr/local/go/pkg/tool/linux_arm64/link: running gcc failed: exit status 1
> gcc: error: unrecognized command-line option '-m64'
> 
> make: *** [build-with-docker] Error 2
> ```
> 
> I think the first error `loadinternal: cannot find runtime/cgo` is because of missing `CGO_ENABLED=1` when building, but the second error `gcc: error: unrecognized command-line option '-m64'` I've got no clue. I bet it's related to cross-platform building, but no matter how I changed the GOOS and GOARCH, it was more or less the same.
> 
> @daniel-hutao please have a look.

It does well at my mac(x86_64), I will see if it can be reproduced after getting the new Mac with m1.

IronCore864:
I am closing this PR at the moment for the ease of issue management and for a clearer view of open PR. Please do not delete the branch, though; let's revisit in January.

Issue #29, opened 2021-12-08
Adding a "planning decision" feature to determine the next action


daniel-hutao:
## Description

Adding a "action calculation" feature to determine the next **action** based on the current **States** of the environment and the input **Config**.

## Describe the Proposal

When the state management function is integrated into the main branch #28 , we need to use the States in the next step, that is, an algorithm is needed to determine the specific actions each time the **dsm** is executed.

The general idea is as follows:

```go
type ActionFunc func(tool *config.Tool) (bool, error)

type Plan struct {
	Changes []*Change
}

type Change struct {
	Tool   *config.Tool
	Action ActionFunc
}

func MakePlan(smgr statemanager.Manager, cfg config.Config) *Plan {...}

// ......
func TestMakePlan() {
	// ......
	plan := MakePlan(smgr, cfg)
	for _, change := range plan.Changes {
		change.Action(change.Tool)
	}
}
```

daniel-hutao:
/assign @daniel-hutao 

I'll make a pr for this feature after #28 is merged.

IronCore864:
Very good. This is an important feature for our v0.1 release.

PR #35, opened 2021-12-09
an "action calculation" feature to determine the next action.

daniel-hutao:
Adding a "action calculation" feature to determine the next action based on the current States of the environment and the input Config.

## Key Points

- [* ] This is a breaking change.

## Description

see #29 

### Related Issues

#29

### DIFF
--- a/cmd/devstream/install.go
+++ b/cmd/devstream/install.go
@@ -27,69 +27,21 @@ func installCMDFunc(cmd *cobra.Command, args []string) {
 	}
 	smgr := statemanager.NewManager(b)
 
-	// this channel will be consume quickly, so its size is no need too big.
-	execResultChan := make(chan plan.ExecutionResult, 1)
-
-	p := plan.MakePlan(smgr, conf)
-	if p.Changes == nil || len(p.Changes) == 0 {
+	p := plan.NewPlan(smgr, conf)
+	if len(p.Changes) == 0 {
 		log.Println("it is nothing to do here")
 		return
 	}
-	go p.Execute(execResultChan)
-	handleResult(smgr, execResultChan)
 
-	log.Printf("===")
-}
-
-func handleResult(smgr statemanager.Manager, execResultChan <-chan plan.ExecutionResult) {
-	var state *statemanager.State
-	// recordState record a single tool's state and return if this state is Succeeded.
-	recordState := func(smgr statemanager.Manager, change *plan.Change) bool {
-		if change.Result.Error != nil {
-			state = statemanager.NewState(
-				change.Tool.Name,
-				change.Tool.Version,
-				[]string{},
-				statemanager.StatusFailed,
-				&statemanager.Operation{
-					Action:   change.ActionName,
-					Time:     change.Result.Time,
-					Metadata: change.Tool.Options,
-				},
-			)
-			smgr.AddState(state)
-			log.Printf("=== plugin %s installation failed ===", change.Tool.Name)
-			return false
-		}
-		state = statemanager.NewState(
-			change.Tool.Name,
-			change.Tool.Version,
-			[]string{},
-			statemanager.StatusRunning,
-			&statemanager.Operation{
-				Action:   change.ActionName,
-				Time:     change.Result.Time,
-				Metadata: change.Tool.Options,
-			},
-		)
-		smgr.AddState(state)
-		log.Printf("=== plugin %s installation done ===", change.Tool.Name)
-		return true
+	errs := p.Execute()
+	if len(errs) == 0 {
+		log.Println("=== all plugins' installation are succeeded ===")
+		return
 	}
 
-	var hasToolFailed bool
-	for result := range execResultChan {
-		for _, change := range result {
-			hasToolFailed = !recordState(smgr, change)
-		}
+	log.Println("=== some errors occurred during plugins installation ===")
+	for _, err := range errs {
+		log.Println(err)
 	}
-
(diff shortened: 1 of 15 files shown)


internal/pkg/plan/plan.go
```diff
@@ -0,0 +1,103 @@
...
+// Change is a wrapper with a single Tool and its Action should be execute.
+type Change struct {
+	Tool       *config.Tool
+	ActionName statemanager.ComponentAction
+	Action     ActionFunc
+}
+
+// MakePlan takes "State Manager" & "Config" then do some calculate and return a Plan.
+// All actions should be execute is included in this Plan.changes.
+func MakePlan(smgr statemanager.Manager, cfg *config.Config) *Plan {
```
IronCore864, line 43: This function is too long and not easily readable.

It's over 50 lines with the comments and doesn't fit on the same screen thus needs scrolling up and down.

Needs to be refactored.

internal/pkg/plan/plan.go
```diff
@@ -0,0 +1,103 @@
...
+						Action:     plugin.Install,
+					})
+				case statemanager.StatusFailed:
+					plan.Changes = append(plan.Changes, &Change{
+						Tool:       &tool,
+						ActionName: statemanager.ActionReinstall,
+						Action:     plugin.Reinstall,
+					})
+				}
+				// The tool hasn't state record at States, so it need to be Installed
```
IronCore864, line 68: Comment should be on the same indentation level as the code.

internal/pkg/plan/plan.go
IronCore864: I think we might not need `statemanager.StatusUninstalled`. If it's uninstalled, we can delete it from the state.

IronCore864: 1. After figuring out the logic, this piece of function can be refactored to something like:

```go
...

for _, tool := range cfg.Tools {
	state := smgr.GetState(tool.Name)
    if state == nil {
		plan.Changes = append(createInstallAction())
		continue
    }

	switch state.Status {
	case statemanager.StatusSucceeded:
		// do nothing
	case statemanager.StatusFailed:
		plan.Changes = append(createReinstallAction())
	case statemanager.StatusUninstalled:
		plan.Changes = append(createInstallAction())
	}
}

for _, state := range states {
	if state.tool not in cfg.Tools {
		plan.Changes = append(createUninstallAction())
	}
}

...
```

What do you think?

2. Even the two loops can be put into two different functions and by giving them proper names, it can be more readable. For example, "generatePlanAccordingToConfig", "removeNoLongerNeededTools".

3. Some functions (like the createXXXAction) can be moved to a separate file, like helper.go or something like that, to make this file shorter and easier to read without much scrolling.
Some files can be moved to

IronCore864 (changes requested):
See comments inline.

internal/pkg/plugin/plugin.go
```diff
@@ -11,9 +11,9 @@ import (
 
 // DevStreamPlugin is a struct, on which install/reinstall/uninstall interfaces are defined.
 type DevStreamPlugin interface {
-	Install(*map[string]interface{})
-	Reinstall(*map[string]interface{})
-	Uninstall(*map[string]interface{})
+	Install(*map[string]interface{}) (bool, error)
```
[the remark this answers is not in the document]
daniel-hutao: Ok

cmd/devstream/install.go
```diff
@@ -20,7 +20,10 @@ func installCMDFunc(cmd *cobra.Command, args []string) {
 	conf := config.LoadConf(configFile)
 	for _, tool := range conf.Tools {
 		log.Printf("=== start to install plugin %s %s ===", tool.Name, tool.Version)
-		plugin.Install(&tool)
+		_, err := plugin.Install(&tool)
```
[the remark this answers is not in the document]
daniel-hutao: You are right, I wanted to do this refactor in the next commit at morning. I'll do it now.

internal/pkg/plan/plan.go
```diff
@@ -0,0 +1,103 @@
+package plan
```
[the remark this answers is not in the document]
daniel-hutao: Maybe you are right, but I think this idea can be discussed more later, and it could not be implemented for now.

internal/pkg/plan/plan.go
```diff
@@ -0,0 +1,103 @@
...
+
+// Plan is an "Actions" plan, it includes all changes should be take with plugins.
+type Plan struct {
+	Changes []*Change
+}
+
+// ExecuteResult holds the execution result with a single plugin.
+// The map's key is a plugin's name.
+// The Result is Succeeded when map's value error is nil.
+type ExecuteResult map[string]error
```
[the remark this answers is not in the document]
daniel-hutao: Good tip！

internal/pkg/plan/plan.go
daniel-hutao: The indentation level is decided by the IDE, I've tried modify the indentation level and it will be revert to this style after "format"

internal/pkg/plan/plan.go
daniel-hutao: Yeah, this function's logic can be improved, I'll have a try to make it more readable and reasonable.

internal/pkg/plan/plan.go
IronCore864: Let me do some tests and try to figure it out. Thanks for the input.

PR #45, opened 2021-12-21
fix: let helm to create namespace for argocd if it does not exist

daniel-hutao:

# Summary

Let helm to create namespace for argocd if it does not exist

## Key Points


daniel-hutao:
I decided to merge this pr first, because the development of #46  will depend on the code here. If you find that there're some problems with this pr, I can to submit a new commit to fix it soon. @IronCore864 @lfbdev

Issue #50, opened 2021-12-22
E2E Testing


IronCore864:
### DTM
- [ ] all dtm params tested

### GitHub Actions Plugin
- [ ] GitHub Actions plugin install from scratch
- [ ] GitHub Actions plugin re-run (install/apply again; reinstall) should not cause any error
- [ ] proper logging messages should be available (for example, xxx pipeline already exists and nothing is changed, etc.)

### ArgoCD Plugin
- [ ] ArgoCD plugin install with existing namespace
- [ ] ArgoCD plugin install without existing namespace
- [ ] ArgoCD plugin reinstall shouldn't change anything
- [ ] proper logging messages should be available

### ArgoCD App Plugin
- [ ] ArgoCD app plugin install
- [ ] ArgoCD app plugin reinstall shouldn't change anything
- [ ] ArgoCD app plugin install without ArgoCD should give a proper warning (CRD doesn't exist; ArgoCD not installed, etc.)

### Automated E2E Testing Investigation

- automated test locally (with minikube, etc.)
- possibility to create a cluster in a container as part of the GitHub actions workflow
- create a small cluster in the cloud with minimum cost (for example, automatically shut down when not used; create/destroy on the fly; etc.)

---

@lfbdev @daniel-hutao, please think about it and add more to the list. We can create a test plan for each plugin in the future based on the list.

daniel-hutao:
We can consider to use a Golang BDD testing framework named Ginkgo.
Github: https://github.com/onsi/ginkgo

I'll have a try for this framework and write a demo for how to use it in DevStream.

daniel-hutao:
All tested cases from me(Live Update):

[ ] xxx

PR #60, opened 2021-12-23
test: adding ginkgo BDD framework for e2e testing & UT

daniel-hutao:

# Summary

Adding ginkgo BDD framework for e2e testing & UT

These are some of my attempts to use ginkgo. Let’s see if it’s ok. If we all agree this framework is good, We could use ginkgo to standardize our test cases later.

### Related Issues

#50


IronCore864:
Great work. We will not merge for now; let's plan an internal knowledge-sharing session at the beginning of January on this topic.

Issue #81, opened 2021-12-28
bug: githubactions "delete" failed with no "sha" error


daniel-hutao:
## Describe the Bug

githubactions "delete" failed with no "sha" error

```
2021/12/28 10:42:42 DELETE https://api.github.com/repos/daniel-hutao/animal/contents/.github/workflows/pr-builder.yml: 422 Invalid request.

"sha" wasn't supplied. []
```

## How to Reproduce

uninstall or reinstall a githubaction

## Screenshots

![image](https://user-images.githubusercontent.com/18692628/147522412-711b3508-de35-420f-845c-c188de216dfa.png)

daniel-hutao:
fixed by #76

lfbdev:
reloved,  [need SHA opt ] (https://github.com/merico-dev/stream/pull/76)

daniel-hutao:
It could be closed now. #86

PR #86, opened 2021-12-28
sha lost in Deleteworkflow

lfbdev:
# Summary

sha lost in Deleteworkflow,  delete option failed

## Key Points

## Description

### Related Issues
#81 

### Current Behavior

### New Behavior

Describe the newly updated behavior, if relevant.

### Screenshots

### Other Information


daniel-hutao (commented):
Some questions:
1. Why the `Branch` parameter needs to be deleted.
2. Why write the temporary version number in ut. Maybe there is no better way to implement it, but doing so means that this piece of code needs to be changed frequently. If you must rely on a version number, I think the version number needs to be externally configurable, or simply not use this test case. Maybe we need to use mock, of course, this is not an urgent matter at the moment.

image

daniel-hutao (approved):
/lgtm

PR #97, opened 2021-12-29
progress bar for download

lfbdev:
# Summary
add progress bar  when executing download plugins

## Key Points

## Description

### Related Issues

#98

### DIFF
--- a/internal/pkg/pluginmanager/pbdownloader.go
+++ b/internal/pkg/pluginmanager/pbdownloader.go
@@ -1,0 +1,116 @@
+// 1. get plugins *.so
+// 2. show progress bar on console
+
+package pluginmanager
+
+import (
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strconv"
+	"time"
+
+	"github.com/cheggaaa/pb"
+)
+
+type PbDownloadClient struct {
+	*http.Client
+}
+
+func NewPbDownloadClient() *PbDownloadClient {
+	dClient := PbDownloadClient{}
+	dClient.Client = http.DefaultClient
+	return &dClient
+}
+
+// download from release assets
+func (pd *PbDownloadClient) download(pluginsDir, pluginFilename, version string) error {
+
+	err := createPathIfNotExists(pluginsDir)
+	if err != nil {
+		return err
+	}
+
+	downloadURL := fmt.Sprintf("%s/v%s/%s", defaultReleaseUrl, version, pluginFilename)
+	tmpName := pluginFilename + ".tmp"
+
+	pd.Timeout = time.Second * 60 * 60
+	resp, err := pd.Get(downloadURL)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode == http.StatusOK {
+		log.Printf("[INFO] downloading: [%s]**\n**", pluginFilename)
+
+		downFile, err := os.Create(filepath.Join(pluginsDir, tmpName))
+		if err != nil {
+			return err
+		}
+
+		defer downFile.Close()
+		// create progress bar when reading response body
+		errSetup := setUpProgressBar(resp, downFile)
+		if errSetup != nil {
+			log.Print(errSetup)
+			return errSetup
+		}
+		log.Printf("[INFO] [%s] download success.", pluginFilename)
+	} else {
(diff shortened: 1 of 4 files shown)


internal/pkg/pluginmanager/pbdownloader.go
```diff
@@ -0,0 +1,119 @@
...
+
+	pd.Timeout = time.Second * 60 * 60
+	resp, err := pd.Get(downloadURL)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode == http.StatusOK {
+		log.Printf("[INFO] downloading: [%s]", pluginFilename)
+		log.Print("\n")
```
daniel-hutao, line 48: Why not just use `log.Printf("[INFO] downloading: [%s]**\n**", pluginFilename)`

internal/pkg/pluginmanager/pbdownloader.go
```diff
@@ -0,0 +1,119 @@
...
+
+		defer downFile.Close()
+		// create progress bar when reading response body
+		errSetup := setUpProgressBar(resp, downFile)
+		if errSetup != nil {
+			log.Print(errSetup)
+			return errSetup
+		}
+
+		log.Print("\n")
```
daniel-hutao, line 63: Same as above, Print("\n") may be unnecessary

internal/pkg/pluginmanager/pbdownloader.go
```diff
@@ -0,0 +1,119 @@
...
+		log.Printf("[INFO] [%s] download success.", pluginFilename)
+	} else {
+		log.Print("\n")
+		log.Printf("[ERROR] [%s] download fail,%s.", pluginFilename, resp.Status)
```
daniel-hutao, line 67: What to do next after the download failed?

internal/pkg/pluginmanager/pbdownloader.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	// rename, tmp file to real file
+	err = os.Rename(
```
daniel-hutao, line 71: Maybe `if err = os.Rename(); err != nil {} ` is more clear

internal/pkg/pluginmanager/pbdownloader.go
```diff
@@ -0,0 +1,119 @@
...
+		log.Print(err)
+		return err
+	}
+	bar.Finish()
+	return nil
+}
+
+func createPathIfNotExists(path string) error {
+	_, err := os.Stat(path)
+	if err == nil {
```
daniel-hutao, line 108: Same here. I'd like to use `if err := func(); err != nil {}`.
There are many places above that have the same problem. Of course, this is not very important, but it is relatively more commonly used.

internal/pkg/pluginmanager/pbdownloader.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil
+	}
+	if os.IsNotExist(err) {
```
daniel-hutao, line 111:
```
if !os.IsNotExist(err) {
 	return err
}
if err := os.Mkdir(path, os.ModePerm); err != nil {
 	return err
}
 return nil

daniel-hutao (changes requested):
That's all. Please make sure that the downloader is tested locally.

internal/pkg/pluginmanager/pbdownloader.go
lfbdev: done

internal/pkg/pluginmanager/pbdownloader.go
lfbdev: done

internal/pkg/pluginmanager/pbdownloader.go
lfbdev: done

internal/pkg/pluginmanager/pbdownloader.go
lfbdev: done

internal/pkg/pluginmanager/pbdownloader.go
lfbdev: done

internal/pkg/pluginmanager/pbdownloader.go
lfbdev: done

IronCore864 (approved):

daniel-hutao (approved):

Issue #104, opened 2022-01-04
dtm init Improvement: Add a Parameter -p (--plugins-dir)


IronCore864:
## Current Situation

Plugins dir: `./plugins/`

State: `./devstream.state`

## New Behaviour

- add a new parameter `-p` or `--plugin-dir`. mkdir if directory doesn't exist.
- default value could be `./.devstream/plugins/`
- state file `./.devstream/devstream.state`

PR #108, opened 2022-01-04
Feat: Add a Parameter -p (--plugins-dir)

daniel-hutao:

# Summary

Add a Parameter -p (--plugins-dir)

### Related Issues

fix: #104

### DIFF
--- a/internal/pkg/pluginmanager/manager.go
+++ b/internal/pkg/pluginmanager/manager.go
@@ -2,33 +2,40 @@ package pluginmanager
 
 import (
 	"errors"
+	"fmt"
 	"log"
 	"os"
 	"path/filepath"
+
+	"github.com/spf13/viper"
 
 	"github.com/merico-dev/stream/internal/pkg/configloader"
 )
 
 func DownloadPlugins(conf *configloader.Config) error {
 	// create plugins dir if not exist
-	pluginsDir := filepath.Join(".", "plugins")
+	pluginDir := viper.GetString("plugin-dir")
+	if pluginDir == "" {
+		return fmt.Errorf("plugins directory should not be \"\"")
+	}
+	log.Printf("prepare to use dir < %s > to hold the plugins", pluginDir)
 
 	// download all plugins that don't exist locally
 	dc := NewDownloadClient()
 
 	for _, tool := range conf.Tools {
 		pluginFileName := configloader.GetPluginFileName(&tool)
-		if _, err := os.Stat(filepath.Join(pluginsDir, pluginFileName)); errors.Is(err, os.ErrNotExist) {
+		if _, err := os.Stat(filepath.Join(pluginDir, pluginFileName)); errors.Is(err, os.ErrNotExist) {
 			// plugin does not exist
-			log.Printf("=== now downloading plugin: %s, version: %s ===", pluginFileName, tool.Version)
-			err := dc.download(pluginsDir, pluginFileName, tool.Version)
+			log.Printf("=== downloading plugin: %s ===", pluginFileName)
+			err := dc.download(pluginDir, pluginFileName, tool.Version)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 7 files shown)


internal/pkg/pluginmanager/manager.go
```diff
@@ -20,15 +27,15 @@ func DownloadPlugins(conf *configloader.Config) error {
 		pluginFileName := configloader.GetPluginFileName(&tool)
 		if _, err := os.Stat(filepath.Join(pluginsDir, pluginFileName)); errors.Is(err, os.ErrNotExist) {
 			// plugin does not exist
-			log.Printf("=== now downloading plugin: %s, version: %s ===", pluginFileName, tool.Version)
+			log.Printf("=== downloading plugin: %s ===", pluginFileName)
```
lfbdev, line 30: if the version should be removed or not

daniel-hutao: I deleted it for the version info has contained in the filename.

lfbdev (approved):

internal/pkg/pluginmanager/manager.go
```diff
@@ -2,16 +2,23 @@ package pluginmanager
...
 
+	"github.com/spf13/viper"
+
 	"github.com/merico-dev/stream/internal/pkg/configloader"
 )
 
 func DownloadPlugins(conf *configloader.Config) error {
 	// create plugins dir if not exist
-	pluginsDir := filepath.Join(".", "plugins")
+	pluginsDir := viper.GetString("plugin-dir")
```
IronCore864, line 17: Let's change it to pluginDir so that it's consistent.

internal/pkg/pluginmanager/manager.go
```diff
@@ -20,15 +27,15 @@ func DownloadPlugins(conf *configloader.Config) error {
...
 		if _, err := os.Stat(filepath.Join(pluginsDir, pluginFileName)); errors.Is(err, os.ErrNotExist) {
 			// plugin does not exist
-			log.Printf("=== now downloading plugin: %s, version: %s ===", pluginFileName, tool.Version)
+			log.Printf("=== downloading plugin: %s ===", pluginFileName)
 			err := dc.download(pluginsDir, pluginFileName, tool.Version)
 			if err != nil {
 				return err
 			}
-			log.Printf("=== plugin: %s, version: %s downloaded ===", pluginFileName, tool.Version)
+			log.Printf("=== plugin: %s has been downloaded ===", pluginFileName)
```
IronCore864, line 35: What is the reason to remove the `version` from the log? Do we need it or not?

internal/pkg/pluginmanager/manager.go
```diff
@@ -20,15 +27,15 @@ func DownloadPlugins(conf *configloader.Config) error {
...
 			continue
 		}
-		log.Printf("=== plugin: %s, version: %s exists ===", pluginFileName, tool.Version)
+		log.Printf("=== plugin: %s exists ===", pluginFileName)
```
IronCore864, line 38: Let's take this chance to make it more user-friendly. I suggest:

"plugin: %s already exists in local plugin-dir %s"

What do you think?

IronCore864 (changes requested):

internal/pkg/pluginmanager/manager.go
daniel-hutao: deleted it for the version info has contained in the filename.

internal/pkg/pluginmanager/manager.go
daniel-hutao: I already printed the plugin-dir above. So it's no need to print the dir info here.

internal/pkg/pluginmanager/manager.go
daniel-hutao: plugin-dir is to be consistent with config-file

image

IronCore864 (approved):

Issue #48, opened 2021-12-22
Change install command to create/apply


daniel-hutao:
## Description

Maybe we need to change the install command to create/apply because the uninstall operation can be triggered during the execution of the install command, so the install command is confusing. 

In addition, the delete operation may require more consideration. If the user accidentally specifies an empty configuration file, after executing the `dtm` command, all installed components will be uninstalled, which may not be what they want. We may need to provide a separate command for the uninstall action, such as the `delete` command, to let users know the real result of their command action.

I consider solving the first problem mentioned above first - rename the install command. 
Then modify the `planmanager` so that `create/apply` does not trigger the uninstall action. 
Finally, add a separate `uninstall/delete` command entrance.

/kind proposal
/assign @daniel-hutao 
/cc @merico-dev/devstream

## Done-Done Check

- [ ] interfaces not changed
- [ ] command change to apply and delete
- [ ] document update (readme, etc)

IronCore864:
Right, I think we only need two commands, apply and uninstall:
- apply makes sure the state and the config are the same (uninstall might be triggered);
- uninstall destroys everything from the state no matter what the config is.

I'm moving this to the backlog for now; let's focus on v0.0.1 first. We will revisit this in 2022 Jan.

daniel-hutao:
Ok, Let's discuss it further in 2022 Jan.

IronCore864:
Delete will be implemented in another ticket. Closing this one for now.

PR #109, opened 2022-01-04
change command 'install' to 'apply'

lfbdev:
# Summary

change dtm command from 'install' to 'apply'

## Description

### Related Issues

#48

### DIFF
--- a/cmd/devstream/install.go
+++ b/cmd/devstream/install.go
@@ -13,9 +13,9 @@ import (
 )
 
 var installCMD = &cobra.Command{
-	Use:   "install",
-	Short: "Install tools defined in config file",
-	Long:  `Install tools defined in config file`,
+	Use:   "apply",
+	Short: "Creates or updates DevOps tools according to DevStream configuration file.",
+	Long:  `Creates or updates DevOps tools according to DevStream configuration file. DevStream will generate and execute a new plan based on the config file and the state file by default.`,
 	Run:   installCMDFunc,
 }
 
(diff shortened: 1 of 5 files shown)


cmd/devstream/install.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 var installCMD = &cobra.Command{
-	Use:   "install",
+	Use:   "apply",
 	Short: "Install tools defined in config file",
 	Long:  `Install tools defined in config file`,
```
IronCore864, line 18: For the short and long descriptions, I think we need a better one for the short and long descriptions for the short and long descriptions.

Example:

Short: Creates or updates DevOps tools according to DevStream configuration file.

Long: Creates or updates DevOps tools according to DevStream configuration file. DevStream will generate and execute a new plan based on the config file and the state file by default.

IronCore864 (changes requested):

cmd/devstream/install.go
IronCore864: Please also make sure "install" is changed to "apply" in all places, such as logging, etc. A grep might help.

lfbdev: > Please also make sure "install" is changed to "apply" in all places, such as logging, etc. A grep might help.

these logs may not be modified:
image

lfbdev: > For the short and long descriptions, I think we need a better one for the short and long descriptions for the short and long descriptions.
> 
> Example:
> 
> Short: Creates or updates DevOps tools according to DevStream configuration file.
> 
> Long: Creates or updates DevOps tools according to DevStream configuration file. DevStream will generate and execute a new plan based on the config file and the state file by default.

modified

daniel-hutao (approved):
/lgtm

IronCore864 (approved):

PR #112, opened 2022-01-05
Refactor: cmd and pluginengine package

### LINKED ISSUE #64: Refactor: cmd and pluginengine Refactor
## Description

Current situation:

- most logic (download plugins, create a backend, read state, create a plan and execute it) all lives in the `cmd/devstream/install.go`; the `internal/pkg/pluginengine` doesn't do much except load the so and execute the corresponding interface
- plan generation code lives in `stream/internal/pkg/planmanager/result.go` which doesn't seem right according to the name
- plan generation calls the engine instead of the other way around

## Describe the Proposal

- move the main logic out of the `cmd/devstream` package and into the `internal/pkg/pluginengine` package
- pluginengine as the main entrance (as shown in the architecture diagram)
- figure out a way so that plan doesn't call engine, but the other way around, so that:
- pluginengine initiates config loading, plugin downloading, plan creating/executing, and state managing

daniel-hutao:

# Summary

Refactor: cmd and pluginengine package

### Related Issues

fix #64

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,9 +3,9 @@ help: ## Display this help.
 
 build: fmt vet ## Build dtm & plugins locally.
 	go mod tidy
-	go build -buildmode=plugin -trimpath -gcflags="all=-N -l" -o plugins/githubactions_0.0.1.so ./cmd/githubactions/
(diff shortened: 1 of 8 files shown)


IronCore864:
Closing now since we already started working on this.

PR #113, opened 2022-01-06
refactor: cmd and plugin engine modules, and fixing some bugs

### LINKED ISSUE #64: Refactor: cmd and pluginengine Refactor

### LINKED ISSUE #102: init and install logic improvement:
What should happen when install is called without calling init first?

Shall we run the init anyway for the user, or only a warning should be given and let the user take control of the whole process?

We can refer to other open-source tools, like Terraform init/apply.

IronCore864:
# Summary

## Key Points

## Description

Build:
- `make build` will make sure the default plugins directory exists (`.devstream`) and will build the plugins to `.devstream/*.so`
- `make fmt` changed to `go install` (starting in Go 1.17, installing executables with go get is deprecated)

Command/Engine:
- init: the logic isn't changed; apply without init will also work.
- apply: the `applyCMDFunc` is hugely reduced, moving the main logic to a new function `pluginengine.Run`
- a new helper file for the engine module is created to better organize the code structure

Configloader:
- viper initialization is moved out of the `initConfig` function (bug: `dtm` without any subcommand would fail because it tries to load the config file)
- viper UnmarshalKey is moved to `configloader.LoadConf` to reduce duplicated code

State/Backend:
- default state file is changed from `.devstream/devstream.state` to `devstream.state`
- local backend will create a new empty file if it doesn't exist. This solves the minor bug that the first time when dtm is executed, there is no state, hence stateMap reading would cause an error output
- fixing a bug that after installation, the state is still "running" instead of "installed"

Plugins:
- default plugins directory: changed from `.devstream/plugins` to `.devstream`.
- argocd plugin: cascading deletion of an app when the Application CRD is deleted (for easier local testing)

**Note: the changes of state/plugins are due to the fact that if using `-p` to specify another directory for plugins, the state file will still live at `.devstream/devstreamstate` but the `.devstream` folder might not exist since it's not created by the `init`.**

Plan:
- `plan.Execute` now is moved to the plugin engine module, so that there are no circular references (engine uses plan.execute, which uses engine.install). Because of this, `change.Action` is removed; only `change.ActionName` is used.
- a `plan_helper` is created to better organize the code, and the `result` is separated with `plan`

Logging:
- removed "===" prefixes and suffixes (will be further improved by introducing another logging module)
- fixed some grammar errors

Test:
- fixed a commented test

### Related Issues

Closes #64 
Closes #102

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ dtm
 
 # .devstream
 .devstream/
+devstream.state
--- a/internal/pkg/backend/local/local.go
+++ b/internal/pkg/backend/local/local.go
@@ -1,13 +1,15 @@
 package local
 
 import (
+	"errors"
 	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
 	"sync"
 )
 
-const DefaultStateFile = ".devstream/devstream.state"
+const DefaultStateFile = "devstream.state"
 
 // Local is a default implement for backend.Backend
 type Local struct {
@@ -21,6 +23,15 @@ func NewLocal(filename string) *Local {
 	if filename == "" {
 		lFile = DefaultStateFile
 	}
+
+	if _, err := os.Stat(lFile); errors.Is(err, os.ErrNotExist) {
+		file, err := os.Create(lFile)
+		if err != nil {
+			log.Fatalf("Creating state file %s failed.", lFile)
+		}
+		defer file.Close()
+	}
+
 	return &Local{
 		filename: lFile,
 	}
--- a/internal/pkg/configloader/config.go
+++ b/internal/pkg/configloader/config.go
@@ -2,6 +2,9 @@ package configloader
 
 import (
 	"fmt"
+	"log"
+
+	"github.com/spf13/viper"
 )
 
 // Config is the struct for loading DevStream configuration YAML files.
@@ -26,6 +29,31 @@ func (t *Tool) DeepCopy() *Tool {
 		retTool.Options[k] = v
 	}
 	return &retTool
+}
+
+// LoadConf reads an input file as a Config struct.
+func LoadConf(fname string) *Config {
+	if fname != "" {
+		viper.SetConfigFile(fname)
+	} else {
+		viper.AddConfigPath(".")
+		viper.SetConfigType("yaml")
+		viper.SetConfigName("config")
+	}
+
+	if err := viper.ReadInConfig(); err != nil {
+		log.Print(err)
+		log.Print("Perhaps you forgot to specify the path of the config file by using the \"-f\" parameter?")
+		log.Fatal("See more help by running \"dtm help\"")
+	}
+
+	var tools = make([]Tool, 0)
+
+	if err := viper.UnmarshalKey("tools", &tools); err != nil {
+		log.Fatal(err)
+	}
+
+	return &Config{Tools: tools}
 }
 
 // GetPluginFileName creates the file name based on the tool's name and version
--- a/internal/pkg/planmanager/result.go
+++ b/internal/pkg/planmanager/result.go
@@ -3,13 +3,11 @@ package planmanager
 import (
 	"log"
 
-	"github.com/merico-dev/stream/internal/pkg/configloader"
-	"github.com/merico-dev/stream/internal/pkg/pluginengine"
 	"github.com/merico-dev/stream/internal/pkg/statemanager"
 )
 
-// handleResult is used to Write the latest StatesMap to Backend.
-func (p *Plan) handleResult(change *Change) error {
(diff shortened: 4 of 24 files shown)


.gitignore
```diff
@@ -27,3 +27,4 @@ dtm
 
 # .devstream
 .devstream/
+devstream.state
```
daniel-hutao, line 30: `devstream.state` will located inside the `.devstream/` directory, so I think it's no need to put the `devstream.state` here.

internal/pkg/backend/local/local.go
```diff
@@ -1,13 +1,15 @@
...
+	"errors"
 	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
 	"sync"
 )
 
-const DefaultStateFile = ".devstream/devstream.state"
+const DefaultStateFile = "devstream.state"
```
daniel-hutao, line 12: Our previous plan was to put this file in the `.devstream/` directory, why did we decide to change it?

internal/pkg/backend/local/local.go
```diff
@@ -21,6 +23,15 @@ func NewLocal(filename string) *Local {
 	if filename == "" {
 		lFile = DefaultStateFile
 	}
+
+	if _, err := os.Stat(lFile); errors.Is(err, os.ErrNotExist) {
```
daniel-hutao, line 27: `Write()` will create the state file if it's not exist

internal/pkg/planmanager/result.go
```diff
@@ -31,67 +29,17 @@ func (p *Plan) handleResult(change *Change) error {
 	// uninstall failed
 	if change.ActionName == statemanager.ActionUninstall && !change.Result.Succeeded {
 		state.Status = statemanager.StatusInstalled
-		log.Printf("=== plugin %s uninstall failed ===", change.Tool.Name)
-		// install or reinstall failed
+		log.Printf("Plugin %s uninstall failed.", change.Tool.Name)
```
daniel-hutao, line 32: Log printing generally starts with lowercase

internal/pkg/configloader/config.go
```diff
@@ -28,6 +31,31 @@ func (t *Tool) DeepCopy() *Tool {
 	return &retTool
 }
 
+// LoadConf reads an input file as a Config struct.
+func LoadConf(fname string) *Config {
```
daniel-hutao, line 35: These codes are duplicated in `cmd/main.go`

daniel-hutao (changes requested):
This pr is a bit too big, it looks a little hard to understand. That's all from me.

.gitignore
IronCore864: `devstream.state` now is moved out of the `.devstream/` folder.

internal/pkg/backend/local/local.go
IronCore864: See the description for the details.

TL;DR: when specifying a plugin directory with "-p", this folder won't be created so there is an error when creating the state file.

internal/pkg/planmanager/result.go
IronCore864: At the moment, we have different formats, which will be improved in the next sprint. Closing this comment for now since we will improve it in the next sprint thoroughly.

Some discussion:

- Log messages are line-oriented, so they may be full sentences. For full sentences, the first letter should be uppercase and it should end with punctuation.
- Au-contraire, errors should start with lowercase letters.

internal/pkg/configloader/config.go
IronCore864: cmd.main sets viper.AutomaticEnv() and binds ENV vars and flags (cmd related);

LoadConf reads the config file (config file related, hence the name "load conf")

IronCore864:
Added some fixes based on code review:

- [ ] state: needs more investigation;
- [x] pluginengine.Do should return an error;
- [x] internal/pkg/planmanager/change.go has duplicated code and comments.

daniel-hutao (approved):

lfbdev (approved):
/lgtm

PR #122, opened 2022-01-11
refactor: only check, no download plugins before apply

### LINKED ISSUE #102: init and install logic improvement:

IronCore864:
# Summary

## Key Points

## Description

This is a breaking change.

### Related Issues

Closes #102 

### Current Behavior

Apply will init first (which will download plugins if they don't exist).

### New Behavior

Apply WILL NOT init first. Apply only checks if needed plugins exist; if not, an error will be returned instead of downloading the plugins for the user.

### Screenshots

Screenshot 2022-01-11 at 11 44 38

### DIFF
--- a/cmd/devstream/apply.go
+++ b/cmd/devstream/apply.go
@@ -22,7 +22,7 @@ func applyCMDFunc(cmd *cobra.Command, args []string) {
 
 	err := pluginengine.Apply(configFile)
 	if err != nil {
-		log.Printf("Apply error: %s. Maybe you forgot to run \"dtm init\" first?", err)
+		log.Printf("Apply error: %s.", err)
 		os.Exit(1)
 	}
 
--- a/cmd/devstream/apply.go
+++ b/cmd/devstream/apply.go
@@ -22,7 +22,7 @@ func applyCMDFunc(cmd *cobra.Command, args []string) {
 
 	err := pluginengine.Apply(configFile)
 	if err != nil {
-		log.Printf("Apply error: %s.", err)
+		log.Printf("Apply error: %s. Maybe you forgot to run \"dtm init\" first?", err)
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/devstream/apply.go
```diff
@@ -22,7 +22,7 @@ func applyCMDFunc(cmd *cobra.Command, args []string) {
 
 	err := pluginengine.Apply(configFile)
 	if err != nil {
-		log.Printf("Apply error: %s.", err)
+		log.Printf("Apply error: %s. Maybe you forgot to run \"dtm init\" first?", err)
```
daniel-hutao, line 25: What if `pluginmanager.CheckLocalPlugins(cfg)` return nil but some error occurred below in `Apply()`?

internal/pkg/pluginmanager/manager.go
```diff
@@ -55,6 +56,27 @@ func DownloadPlugins(conf *configloader.Config) error {
...
+
+	log.Printf("Using dir <%s> to store plugins.", pluginDir)
+
+	for _, tool := range conf.Tools {
+		pluginFileName := configloader.GetPluginFileName(&tool)
+		if _, err := os.Stat(filepath.Join(pluginDir, pluginFileName)); errors.Is(err, os.ErrNotExist) {
+			if err != nil {
+				return err
+			}
+			return fmt.Errorf("plugin %s directory doesn't exist", tool.Name)
```
daniel-hutao, line 73: delete the word "directory"

daniel-hutao (changes requested):

cmd/devstream/apply.go
IronCore864: Good point. This error message should be put in the `Apply()` func right after checking plugins. Will move it.

daniel-hutao (approved):
lgtm

lfbdev (approved):
/lgtm

PR #124, opened 2022-01-11
Feat: Automated Local E2E Testing - Test 0 to 1

daniel-hutao:

# Summary

Automated Local E2E Testing - Test 0 to 1


daniel-hutao:
/cc @IronCore864 @lfbdev Since there are about 200 lines of changes in this pr after the smoke framework is implemented, I want to merge it first; the health check function of the remaining three plugins will be implemented in a separate commit.

Please take a review at this pr.

daniel-hutao:
I deleted the smoke/argocd/argocd.go just now. already rebase & force pushed

lfbdev (approved):

internal/pkg/pluginengine/check.go
```diff
@@ -0,0 +1,30 @@
+package pluginengine
+
+import (
+	"log"
+
+	"github.com/merico-dev/stream/internal/pkg/configloader"
+)
+
+// CheckHealthy return true while all tools are healthy
```
IronCore864, line 9: CheckHealthy return true while all tools are healthy ->

CheckHealthy return**s** true if all tools are healthy

IronCore864 (approved):

internal/pkg/pluginengine/check.go
daniel-hutao: get it

PR #129, opened 2022-01-14
refactor: remove status and operations from state

IronCore864:
# Summary

## Key Points

## Description

- "last operation", and "status" are removed from the state
- refactor the code to adopt to the change
- update tests


cmd/devstream/delete.go
```diff
@@ -18,6 +19,10 @@ DevStream will delete everything defined in the config file, regardless of the s
 
 func deleteCMDFunc(cmd *cobra.Command, args []string) {
 	log.Println("Delete started.")
-	pluginengine.Delete(configFile)
+	err := pluginengine.Delete(configFile)
+	if err != nil {
```
lfbdev, line 23: maybe the following format is better: 
if err := func(); err != nil {
 //todo
}

internal/pkg/planmanager/result.go
```diff
@@ -8,38 +8,26 @@ import (
...
-
-	// uninstall failed
-	if change.ActionName == statemanager.ActionUninstall && !change.Result.Succeeded {
-		state.Status = statemanager.StatusInstalled
-		log.Printf("Plugin %s uninstall failed.", change.Tool.Name)
-	} else if !change.Result.Succeeded {
-		// install or reinstall failed
-		state.Status = statemanager.StatusFailed
-		log.Printf("Plugin %s (re)install failed.", change.Tool.Name)
+	if change.Result.Succeeded {
```
daniel-hutao, line 11: if !change.Result.Succeeded{
    xxx
    return
}
xxxx

internal/pkg/pluginengine/delete.go
```diff
@@ -30,17 +30,18 @@ func Delete(fname string) {
...
 		log.Println("Nothing needs to be deleted. There is nothing to do.")
-		return
+		return nil
 	}
 
 	errsMap := execute(p)
-	if len(errsMap) == 0 {
-		log.Println("All plugins deleted successfully.")
-		return
+	if len(errsMap) != 0 {
```
daniel-hutao, line 37: if len(errsMap) == 0 {
 	log.Println("All plugins deleted successfully.")
 	return nil
}

for k, e := range errsMap {
	log.Printf("%s -> %s", k, e)
}
	return errors.New("some error(s) occurred during plugins delete process")
 }

daniel-hutao (changes requested):
That's all from me.

internal/pkg/pluginengine/delete.go
IronCore864: This function is a bit long, so putting the "normal" return result at the end makes it easier to read. Otherwise, the reader might need to dig in the 30-line of code to see where the normal return is.

lfbdev (approved):
/lgtm

daniel-hutao (approved):

Issue #79, opened 2021-12-28
proposal: githubactions supports multi-repo


daniel-hutao:
## Description

githubactions supports multi-repo

PR #135, opened 2022-01-17
feat: plugin supports multiple instances

IronCore864:
# Summary

## Key Points

## Description

- `viper` is changed to `yaml.v3` for config loading. This is because `viper` loads the object into `map[interface{}]interface{}` instead of `map[string]interface{}` (the latter of which is the state).
- `reflect.DeepEqual` is changed to `cmp.Equal`, which is intended to be a more powerful and safer alternative for comparing whether two values are semantically equal.
- bug fix: fix a bug that the argocdapp plugin only deletes the temp file in Uninstall but not in Install
- argocd app: timeout increased to 5 min. This is because, it seems that the go-helm-client doesn't do the upgrade in a `helm upgrade --install` fashion and will generate an error if the first install fails.
- some minor refactor to make the Go code more idiomatic.

### DIFF
--- a/examples/config.yaml
+++ b/examples/config.yaml
@@ -21,6 +21,17 @@ tools:
         coverage:
           enable: False
           profile: "-race -coverprofile=coverage.out -covermode=atomic"
+- name: dtm-test1-ci
+  plugin:
+    kind: githubactions
+    version: 0.0.1
+  options:
+    owner: ironcore864
+    repo: dtm-test1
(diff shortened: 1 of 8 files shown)


daniel-hutao (approved):
/lgtm

lfbdev (approved):
/lgtm

IronCore864:
Closes #79

PR #142, opened 2022-01-19
Feat: Adding config validation for dtm

daniel-hutao:

# Summary

Adding config validation for dtm.
(Options validations with all plugins will be implemented in another pr)

### DIFF
--- a/internal/pkg/configloader/config.go
+++ b/internal/pkg/configloader/config.go
@@ -15,6 +15,11 @@ type Config struct {
 
 // Tool is the struct for one section of the DevStream configuration file.
 type Tool struct {
+	// RFC 1123 - DNS Subdomain Names style
+	// contain no more than 253 characters
+	// contain only lowercase alphanumeric characters, '-' or '.'
+	// start with an alphanumeric character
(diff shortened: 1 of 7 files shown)


internal/pkg/configloader/config.go
```diff
@@ -47,14 +52,24 @@ func LoadConf(fname string) *Config {
...
+
+	errs := config.Validate()
+	// valid config
+	if len(errs) == 0 {
+		return &config
 	}
 
-	return &config
+	// invalid config
+	for _, e := range errs {
```
IronCore864, line 69: I think it's better to return &config as the last line of this function; it's easier for readers to get the whole picture without reading every line of the code. In the current structure, the reader has to dig to find out what is the normal return result.

IronCore864 (approved):

internal/pkg/configloader/config.go
daniel-hutao: solved

lfbdev (approved):
/lgtm

IronCore864 (approved):

PR #146, opened 2022-01-19
Feat: logging improvement

lfbdev:
# Summary

logging improvement, using a advanced CLI log framework , showing color and level

## Key Points

## Description

logging improvement, using a advanced CLI log framework , showing color and level

### Related Issues
#148 

### Current Behavior

### New Behavior

pigxN1UZuA

VQh5RwqTHt

### Screenshots

### Other Information


internal/pkg/util/helm/helm.go
```diff
@@ -2,10 +2,11 @@ package helm
 
 import (
 	"context"
-	"log"
 	"strings"
 	"time"
 
+	"github.com/merico-dev/stream/internal/pkg/util/log"
```
daniel-hutao, line 8: helmclient "github.com/mittwald/go-helm-client"
"helm.sh/helm/v3/pkg/repo"

"github.com/merico-dev/stream/internal/pkg/util/log"

IronCore864:
1. Update Status in the Same Line of Log

We could refer to `kind`'s implementation, which requires significant change. We could also implement it ourselves, but it is remaking an existing wheel.

That leaves us two options: either do it the `kind` way, or we don't do it. We are quite close to v0.1 release, so let's decide not to update status in the same line of log. We keep it the way as it already is.

2. Choosing the Right Logging Package/Framework

There are 2 major issues with cligger:
- while it can print different levels of logs, there is no switch to turn off a certain level log (like debug; and it doesn't have debug log level)
- there isn't many projects that use it

So, it's better we don't use it and choose a more popular one.

The most common ones probably are glog and logrus; and there are [more](https://github.com/avelino/awesome-go#logging).

It seems glog doesn't have debug level log (correct me if I'm wrong); so let's stick with logrus (or a even more widely used one).

3. The Logging of the Plugins

Plugins should also keep the same log format as the core.
目前看来只有core的日志是用框架打的，插件的不是。插件的日志风格应该跟core一致

PR #167, opened 2022-01-26
Fix: trello duplicated name

daniel-hutao:

# Summary

Fix: trello duplicated name


lfbdev (approved):

internal/pkg/plugin/trellogithub/trellogithub.go
```diff
@@ -257,33 +257,46 @@ type TrelloItemId struct {
...
 
 	return &TrelloItemId{
 		boardId:     board.ID,
-		todoListId:  todo.ID,
-		doingListId: doing.ID,
-		doneListId:  done.ID,
+		todoListId:  todo,
+		doingListId: doing,
+		doneListId:  done,
 	}, nil
```
IronCore864, line 300: This file is already big enough. Please split it up into smaller ones.

internal/pkg/util/trello/trello.go
```diff
@@ -41,4 +41,5 @@ func (c *Client) CreateList(board *trello.Board, listName string) (*trello.List,
 		return nil, fmt.Errorf("listName name can't be empty")
 	}
 	return c.Client.CreateList(board, listName, trello.Defaults())
+
```
IronCore864, line 44: Unnecessary empty line.

daniel-hutao:
"This file is already big enough. Please split it up into smaller ones." - @IronCore864 

This pr has been merged the code of #168 , so it may be necessary to merge the #168 first, then this pr will be clear. In addition, I'd like to implement "split it up into smaller ones" in a separate pr, in order to avoid modifying too much content at one time. @IronCore864

internal/pkg/plugin/trellogithub/trellogithub.go
```diff
@@ -257,33 +257,34 @@ type TrelloItemId struct {
...
+	if len(lists) != 3 {
+		log.Errorf("Unknown lists format: len==%d", len(lists))
+		return nil, fmt.Errorf("unknown lists format: len==%d", len(lists))
 	}
 
-	done, err := c.CreateList(board, "DONE")
-	if err != nil {
-		return nil, err
-	}
+	todo := lists[0].ID
```
IronCore864, line 277: Please add a comment here for future reference. Document why we process it by order. In case Trello changes its API or order in the future, we can quickly locate the bug.

IronCore864 (approved):

PR #175, opened 2022-01-27
add  debug level switch

lfbdev:
# Summary

add debug level switch in CMD: 
./dtm-darwin-arm64 apply -f examples/config.yaml -d 
(open debug level)

### Related Issues

#176 
image

### Other Information

### DIFF
--- a/cmd/devstream/apply.go
+++ b/cmd/devstream/apply.go
@@ -19,6 +19,7 @@ DevStream will generate and execute a new plan based on the config file and the 
 }
 
 func applyCMDFunc(cmd *cobra.Command, args []string) {
+	initLogConf()
 	log.Info("Apply started.")
 
 	if err := pluginengine.Apply(configFile, continueDirectly); err != nil {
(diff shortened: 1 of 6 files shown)


daniel-hutao (changes requested):
,

cmd/devstream/main.go
```diff
@@ -68,6 +70,14 @@ func initConfig() {
 	}
 }
 
+func initLogConf() {
```
daniel-hutao, line 73: I think initLog() is more clear.

daniel-hutao (changes requested):

cmd/devstream/apply.go
```diff
@@ -19,6 +19,7 @@ DevStream will generate and execute a new plan based on the config file and the
 }
 
 func applyCMDFunc(cmd *cobra.Command, args []string) {
+	initLogConf()
```
daniel-hutao, line 22: There should be a more appropriate place to add this code instead of calling it at the entry point of all subcommands

daniel-hutao (changes requested):

PR #178, opened 2022-01-27
Fix: deleting state failed

daniel-hutao:

# Summary

- Fix state delete failed error
- Fix Statesmap issue
- Fix Concurrentmap issue
- Add UT for Concurrentmap and Statesmap

### DIFF
--- a/internal/pkg/configloader/config.go
+++ b/internal/pkg/configloader/config.go
@@ -5,9 +5,9 @@ import (
 	"io/ioutil"
 	"runtime"
 
-	"github.com/merico-dev/stream/internal/pkg/log"
-
 	"gopkg.in/yaml.v3"
+
+	"github.com/merico-dev/stream/internal/pkg/log"
 )
 
 var (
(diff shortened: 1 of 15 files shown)


internal/pkg/planmanager/change.go
```diff
@@ -21,6 +21,11 @@ type ChangeResult struct {
 	Time      string
 }
 
+func (c *Change) String() string {
+	return fmt.Sprintf("\n{\n  ActionName: %s,\n  Tool: {Name: %s, Plugin: {Kind: %s, Version: %s}}\n}",
```
lfbdev, line 25: not log but fmt？

daniel-hutao: return string

lfbdev (approved):

internal/pkg/configloader/config.go
```diff
@@ -5,14 +5,14 @@ import (
...
-
 	"gopkg.in/yaml.v3"
+
+	"github.com/merico-dev/stream/internal/pkg/log"
 )
 
 var (
-	GOOS   string = runtime.GOOS
-	GOARCH string = runtime.GOARCH
+	GOOS   = runtime.GOOS
```
IronCore864, line 14: Please keep the original typed string constant, instead of using an untyped string constant. Unless there is a clear need for this.

Read more here: https://go.dev/blog/constants (section String constants)

IronCore864 (changes requested):

internal/pkg/configloader/config.go
daniel-hutao: right

daniel-hutao:
@IronCore864 Please take a review again.

IronCore864 (approved):

PR #184, opened 2022-02-09
refactor: splitting github actions plugin into different ones accordi…

IronCore864:
# Summary

## Key Points

## Description

`githubactions` plugin is split into 3 plugins (by programming language):
- `githubactions-golang`
- `githubactions-python`
- `githubactions-nodejs`

There will be duplicated code, but it's OK since they are three different plugins now. They may grow into different directions (at the moment, the Golang one already differs from the other two since it may also create repo secrets)

Creating GitHub client is updated to use `pkg/util/github`.

For Golang, the pipeline is split into two (one PR builder, one main builder; the reason being some step is only required in the PR builder.)

Improved secrets handling during reinstall/uninstall for Golang.


daniel-hutao (approved):

lfbdev (approved):
/lgtm

Issue #185, opened 2022-02-10
githubactions default config and behaviour update


lfbdev:
## Description
There are 3 issues.

The “jobs” section doesn’t do anything. We should remove it. From:

- name: golang-demo-app
  plugin:
    kind: githubactions
    version: 0.0.2
  options:
    owner: ironcore864
    repo: go-hello-http
    language:
      name: go
      version: "1.17"
    branch: master
    jobs:
      build:
        enable: True
        command: "go build ./..."
      test:
        enable: True
        command: "go test ./..."
        coverage:
          enable: False
          profile: "-race -covermode=atomic"
          output: "coverage.out"
      docker:
        enable: True
        repo: go-hello-http

to:

- name: golang-demo-app
  plugin:
    kind: githubactions
    version: 0.0.2
  options:
    owner: ironcore864
    repo: go-hello-http
    language:
      name: go
      version: "1.17"
    branch: master
    build:
      enable: True
      command: "go build ./..."
    test:
      enable: True
      command: "go test ./..."
      coverageEnabled: True
      coverageProfile: "-race -covermode=atomic"
      coverageOutput: "coverage.out"
    docker:
      enable: True
      repo: go-hello-http

It’s much easier to read.

2. The config is too long with many standard commands. We should change it further to:

- name: golang-demo-app
  plugin:
    kind: githubactions
    version: 0.0.2
  options:
    owner: ironcore864
    repo: go-hello-http
    language:
      name: go
      version: "1.17"
    branch: master
    test:
      coverageEnabled: True
    docker:
      enable: True

And it should still generate exactly the same result as the previous config.

I.E., if “build” doesn’t exist, we enable it by default; if “build command” doesn’t exist, we set it as “go build ./..." as default.

However, if some config does exist, it should override the default config.

3. githubactions plugin didn’t use our util.github package.

We should try to use the same package.

util.github should also be refactored (TODOs in the code, and another ticket)

PR #186, opened 2022-02-10
remove 'job' item, support default config

lfbdev:
# Summary
1. Remove “jobs” section , as it doesn’t do anything.
2. Support default config if there is no config

## Key Points

### Related Issues
#185 

### Other Information

### DIFF
--- a/examples/config.yaml
+++ b/examples/config.yaml
@@ -21,20 +21,11 @@ tools:
       name: go
       version: "1.17"
     branch: main
-    jobs:
-      build:
-        enable: True
-        command: "go build ./..."
-      test:
-        enable: True
-        command: "go test ./..."
-        coverage:
-          enable: True
-          profile: "-race -covermode=atomic"
-          output: "coverage.out"
-      docker:
-        enable: True
-        repo: golang-demo
+    test:
+      coverage:
+        enable: "True"
+    docker:
+      enable: "True"
 - name: trello-github-integ-golang-demo
   plugin:
     kind: trello-github-integ
--- a/examples/config.yaml
+++ b/examples/config.yaml
@@ -21,11 +21,14 @@ tools:
       name: go
       version: "1.17"
     branch: main
+    build:
+      enable: True
     test:
+      enable: True
       coverage:
-        enable: "True"
+        enable: True
     docker:
-      enable: "True"
+      enable: True
 - name: trello-github-integ-golang-demo
   plugin:
     kind: trello-github-integ
--- a/internal/pkg/plugin/githubactions/golang/install.go
+++ b/internal/pkg/plugin/githubactions/golang/install.go
@@ -27,7 +27,7 @@ func Install(options *map[string]interface{}) (bool, error) {
 	log.Infof("Language is: %s.", ga.GetLanguage(opt.Language))
 
 	// if docker is enabled, create repo secrets for DOCKERHUB_USERNAME and DOCKERHUB_TOKEN
-	if opt.Docker.Enable == "True" {
+	if opt.Docker.Enable {
 		if err := gitHubClient.AddRepoSecret("DOCKERHUB_USERNAME", viper.GetString("dockerhub_username")); err != nil {
 			return false, err
 		}
--- a/internal/pkg/plugin/githubactions/golang/install.go
+++ b/internal/pkg/plugin/githubactions/golang/install.go
@@ -1,10 +1,11 @@
 package golang
 
 import (
+	"github.com/spf13/viper"
+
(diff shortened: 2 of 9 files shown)


internal/pkg/plugin/githubactions/golang/install.go
```diff
@@ -1,11 +1,10 @@
 package golang
 
 import (
-	"github.com/spf13/viper"
-
 	"github.com/merico-dev/stream/internal/pkg/log"
 	ga "github.com/merico-dev/stream/internal/pkg/plugin/githubactions"
 	"github.com/merico-dev/stream/pkg/util/github"
+	"github.com/spf13/viper"
```
daniel-hutao, line 7: format error

internal/pkg/plugin/githubactions/golang/install.go
```diff
@@ -28,7 +27,7 @@ func Install(options *map[string]interface{}) (bool, error) {
 	log.Infof("Language is: %s.", ga.GetLanguage(opt.Language))
 
 	// if docker is enabled, create repo secrets for DOCKERHUB_USERNAME and DOCKERHUB_TOKEN
-	if opt.Jobs.Docker.Enable {
+	if opt.Docker.Enable == "True" {
```
daniel-hutao, line 30: `strings.ToLower(opt.Docker.Enable) == "true"` is better

internal/pkg/plugin/githubactions/golang/ishealthy.go
```diff
@@ -40,7 +40,7 @@ func IsHealthy(options *map[string]interface{}) (bool, error) {
 	}
 
 	// if docker is enabled, verify if secrets DOCKERHUB_USERNAME and DOCKERHUB_TOKEN are deleted
-	if opt.Jobs.Docker.Enable {
+	if opt.Docker.Enable == "True" {
```
daniel-hutao, line 43: Same as above

internal/pkg/plugin/githubactions/golang/reinstall.go
```diff
@@ -27,7 +27,7 @@ func Reinstall(options *map[string]interface{}) (bool, error) {
 
 	log.Infof("language is %s", ga.GetLanguage(opt.Language))
 
-	if opt.Jobs.Docker.Enable {
+	if opt.Docker.Enable == "True" {
```
daniel-hutao, line 30: Same as above

internal/pkg/plugin/githubactions/golang/uninstall.go
```diff
@@ -26,7 +26,7 @@ func Uninstall(options *map[string]interface{}) (bool, error) {
 	log.Infof("language is %s", ga.GetLanguage(opt.Language))
 
 	// if docker is enabled, delete repo secrets DOCKERHUB_USERNAME and DOCKERHUB_TOKEN
-	if opt.Jobs.Docker.Enable {
+	if opt.Docker.Enable == "True" {
```
daniel-hutao, line 29: Same as above

examples/config.yaml
```diff
@@ -21,20 +21,11 @@ tools:
       name: go
       version: "1.17"
     branch: main
-    jobs:
-      build:
```
daniel-hutao, line 25: if “build” doesn’t exist, we enable it by default; if “build command” doesn’t exist, we set it as “go build ./..." as default.

---

I think `build` needs to be retained, otherwise users will not know how to fill in when they need to specify the build command. The corresponding code logic behind may also need to be adjusted accordingly

daniel-hutao (changes requested):

internal/pkg/plugin/githubactions/golang/jobs.go
```diff
@@ -1,13 +1,29 @@
...
 // Jobs is the struct for github actions custom config.
-type Jobs struct {
-	Build  Build
-	Test   Test
-	Docker Docker
+//type Jobs struct {
+//	Build  Build
+//	Test   Test
+//	Docker Docker
+//}
```
IronCore864, line 8: Dead code; delete instead of comment it out.

examples/config.yaml
```diff
@@ -21,20 +21,14 @@ tools:
...
-        command: "go test ./..."
-        coverage:
-          enable: True
-          profile: "-race -covermode=atomic"
-          output: "coverage.out"
-      docker:
-        enable: True
-        repo: golang-demo
+    docker:
+      enable: True
```
daniel-hutao, line 31: I think setting it to False by default is better

daniel-hutao (approved):
lgtm

IronCore864 (approved):

PR #202, opened 2022-02-16
Feat: Refactor githubactions trello scaffolding

lfbdev:
# Summary

Refactor githubactions/trello/scaffolding for new interface


cmd/reposcaffolding/github/golang/main.go
```diff
@@ -13,23 +13,23 @@ const NAME = "github-repo-scaffolding-golang"
 type Plugin string
 
 // Install implements the installation of the github-repo-scaffolding-golang.
```
IronCore864, line 15: Update comment accordingly as well; same below.

cmd/trellogithub/main.go
```diff
@@ -12,23 +12,23 @@ const NAME = "trellogithub"
 type Plugin string
 
 // Install implements the installation of some trello-github-integ workflows.
```
IronCore864, line 14: Update comment accordingly as well; same below.

internal/pkg/plugin/githubactions/golang/update.go
```diff
@@ -8,11 +8,11 @@ import (
 	"github.com/merico-dev/stream/pkg/util/github"
 )
 
-// Reinstall remove and set up GitHub Actions workflows.
-func Reinstall(options *map[string]interface{}) (bool, error) {
+// Update remove and set up GitHub Actions workflows.
+func Update(options *map[string]interface{}) (map[string]interface{}, error) {
```
IronCore864, line 12: Is it possible to call "Delete" first then "Create"? If so, the "Update" implementation would have only 2 single lines of code which is easier to read and has less duplicated code.

IronCore864: Same for other plugins. Maybe it's not possible; just to think about it.

internal/pkg/plugin/trellogithub/github.go
```diff
@@ -183,3 +183,28 @@ func (gi *TrelloGithub) AddTrelloIdSecret(trelloId *TrelloItemId) error {
 
 	return nil
 }
+
+func (gi *TrelloGithub) GetWorkflowState() (map[string]interface{}, error) {
```
IronCore864, line 187: Same as the other one, maybe a better naming could improve the readability of the code.

internal/pkg/plugin/trellogithub/read.go
```diff
@@ -28,8 +29,13 @@ func IsHealthy(options *map[string]interface{}) (bool, error) {
...
+		return nil, nil
 	}
 
-	return true, nil
+	c, err := trello.NewClient()
+	if err != nil {
+		return nil, err
+	}
+
+	return c.GetBoard()
```
IronCore864, line 40: "Read" returns state+error, but "GetGoard" doesn't really seem to be related with "state". Maybe we can create a better name.

Some examples (not good but clearer):

1. GetBoardsAndBuildState?
2. boards = getBoards(); return buildState(boards), nil

pkg/util/github/workflow.go
```diff
@@ -141,3 +141,28 @@ func (c *Client) VerifyWorkflows(workflows []*Workflow) (map[string]error, error
 	}
 	return retMap, nil
 }
+
+func (c *Client) GetWorkflowState() (map[string]interface{}, error) {
```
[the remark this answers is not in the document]
IronCore864: Maybe we don't need to put it under pkg/util, because the return value (which is a state) is only related to some plugins, not all.

Or, we can split this function into two, one gets the workflow, the other creates the state based on the workflow. Getting the workflow is more generic thus fits in the pkg/util; but building state isn't.

IronCore864 (changes requested):

internal/pkg/plugin/githubactions/golang/update.go
lfbdev: because "Delete"  and "Create" are both include  "parseAndValidateOptions(options)" and "NewClient" operations, they are repetitive. so I merge the public operations into "Update", instead of using "Delete" + "Create"

PR #241, opened 2022-02-28
Fix: `dtm init xxx` created the state file

### LINKED ISSUE #240: devstream.state file created even when "apply" isn't executed.
## Describe the Bug

When we run `dtm init`, `dtm version`, or simply `dtm`, the `devstream.state` file is created.

## Expected Behavior

We should only create the state file when we run `dtm apply`; for other commands like `version`, `init` that don't rely on the state file, we don't need to create the file prematurely.

daniel-hutao:

# Summary

Fix: `dtm init xxx` created the state file

## Description

### Related Issues

close #240

### DIFF
--- a/internal/pkg/backend/backend.go
+++ b/internal/pkg/backend/backend.go
@@ -4,21 +4,12 @@ import (
 	"fmt"
 
 	"github.com/merico-dev/stream/internal/pkg/backend/local"
+	"github.com/merico-dev/stream/pkg/util/log"
 )
 
-type BackendType string
+type Type string
 
-const BackendLocal BackendType = "local"
-
-var backends map[BackendType]Backend
-
-func init() {
-	if backends == nil {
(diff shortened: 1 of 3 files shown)


internal/pkg/backend/backend.go
```diff
@@ -4,21 +4,12 @@ import (
 	"fmt"
 
 	"github.com/merico-dev/stream/internal/pkg/backend/local"
+	"github.com/merico-dev/stream/pkg/util/log"
 )
 
-type BackendType string
+type Type string
```
daniel-hutao, line 10: the `BackendType` and `BackendLocal` below start with the package name isn't a good name style. Because the users will write the code like: `backend.BackendType` or backend.BackendLocal`. Obviously the `backend.Type & backend.Local` is better.

lfbdev (approved):

IronCore864 (approved):

PR #256, opened 2022-03-03
Feat: add jira-github-integ plugin

lfbdev:
# Summary

Add  jira-github-integ plugin

- when create a issue from github, it will be created in jira todo
-  when comment a issue from github, it will be commented in jira 
-  when edit a issue from github, it will be moved to in progress in jira 
-  when close a issue from github, it will be moved to done in jira 
-  when reopen a issue from github, it will be moved to todo in jira 

### Other Information

### DIFF
--- a/docs/plugins/jira-github-integ_plugin.md
+++ b/docs/plugins/jira-github-integ_plugin.md
@@ -1,0 +1,48 @@
+## 1 `jira-github-integ` Plugin
+
+This plugin integrates Jira with your GitHub repo.
+
+## 2 Usage:
+
+_Please confirm the preconditions:_
+
+- Jira language must be English
+- There should be an existing Jira project
+
+_This plugin depends on the following two environment variables:_
+
+- JIRA_API_TOKEN
+- GITHUB_TOKEN
+
+Set the values accordingly before using this plugin.
+
+
+To create a Jira API token, see [here](https://support.atlassian.com/atlassian-account/docs/manage-api-tokens-for-your-atlassian-account/).
+
+```yaml
+tools:
+- name: jira-github-integ-default
+  # plugin profile
+  plugin:
+    # kind of this plugin
+    kind: jira-github-integ
+    # version of the plugin
+    version: 0.3.0
+  # options for the plugin
+  # checkout the version from the GitHub releases
+  options:
+    # the repo's owner
+    owner: lfbdev
+    # the repo where you'd like to setup GitHub Actions
+    repo: opendeveloper
+    # "base url: https://id.atlassian.net"
+    jiraBaseUrl: https://merico.atlassian.net
+    # "need real user email in cloud Jira"
+    jiraUserEmail: fangbao.li@merico.dev
+    # "get it from project url, like 'HEAP' from https://merico.atlassian.net/jira/software/projects/HEAP/pages"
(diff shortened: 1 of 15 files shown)


internal/pkg/plugin/jiragithub/workflow.go
```diff
@@ -0,0 +1,14 @@
+package jiragithub
+
+import (
+	github "github.com/merico-dev/stream/pkg/util/github"
+)
+
+const (
+	CommitMessage   = "jira-github-integ Actions workflow, created by DevStream"
+	BuilderFileName = "jira-github-integ.yml"
```
IronCore864, line 9: The "name" of the workflow (as defined in the template) is "name: Jira Github integration Builder", but the file name is "jira-github-integ.yml". I think we need better consistency here.

internal/pkg/plugin/jiragithub/create.go
```diff
@@ -0,0 +1,36 @@
...
+		if err := gitHubClient.AddWorkflow(w, opt.Branch); err != nil {
+			return nil, err
+		}
+	}
+
+	if err := setRepoSecrets(gitHubClient); err != nil {
+		return nil, err
+	}
+
+	return ga.BuildState(opt.Owner, opt.Repo), nil
```
IronCore864, line 35: In this case, `jira-github-integ` and `githubactions` plugins are both maintained by us, but in theory, DevStream plugins will be created and maintained by different committers.

So, it's customary that plugin A should not depend on plugin B's code.

Here, "BuildState" actually is part of the githubactions' internal (private) code.

If similarities are found between different plugins, we could theoretically create a public module under `/pkg`.

In this specific situation, I don't think this is the right think to do, though, because it's building "state" that specifically belongs to one plugin. Although the state are exactly the same for these two plugins for now, they might differ in the future.

IronCore864 (changes requested):

internal/pkg/plugin/jiragithub/create.go
```diff
@@ -0,0 +1,36 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	ghOptions := &github.Option{
+		Owner:    opt.Owner,
+		Repo:     opt.Repo,
+		NeedAuth: true,
+	}
+	gitHubClient, err := github.NewClient(ghOptions)
```
daniel-hutao, line 20: I recommend using `ghClient` as we've named it in other packages.

docs/plugins/jira-github-integ_plugin.md
```diff
@@ -0,0 +1,40 @@
...
+This plugin integrates jira with your GitHub repo.
+
+## 2 Usage:
+
+_This plugin depends on the following two environment variables:_
+
+- JIRA_BASE_URL
+- JIRA_USER_EMAIL
+- JIRA_API_TOKEN
+- JIRA_PROJECT_KEY
```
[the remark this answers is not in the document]
daniel-hutao: +1

daniel-hutao:
> when editing an issue from GitHub, it will be moved to in progress in Jira

I don't think this logic makes sense; perhaps `in progress` can be judged by comments like `/assign`

daniel-hutao (approved):

Issue #263, opened 2022-03-03
:bug: apply and create git repo success, then delete from github, apply again and fail


JaneLiuL:
## Describe the Bug
below is my config, the first time i run apply -f ./config.yml
and check the repo on github, and i delete the new repo
then i run again apply -f ./config.yml , it fail as below
<img width="936" alt="image" src="https://user-images.githubusercontent.com/56822274/156546397-ecb85887-9f1d-414f-abab-d5ae997e218a.png">

```
---
tools:
# name of the instance with github-repo-scaffolding-golang
- name: golang-demo
  plugin:
    kind: github-repo-scaffolding-golang
    version: 0.2.0
  options:
    owner: janeliul
    # the repo which you'd like to create
    repo: test-demo
    # the branch of the repo you'd like to hold the code
    branch: main
    image_repo: xxxx/golang-demo
- name: golang-demo-actions
  plugin:
    kind: githubactions-golang
    version: 0.2.0
  options:
    owner: janeliul
    # the repo where you'd like to setup GitHub Actions
    repo: test-demo
    # programming language specific settings
    language:
      name: go
      # version of the language
      version: "1.17"
    # main branch of the repo (to which branch the plugin will submit the workflows)
    branch: main
    build:
      # default to true
      enable: True
      # build command, OPTIONAL, the given value below is default value
      command: "go build ./..."
    test:
      # default to false
      enable: False
      # test command, OPTIONAL, the given value below is default value
      command: "go test ./..."
      coverage:
        # default to false
        enable: False
        # go test profile subcommand, OPTIONAL, the given value below is default value
        profile: "-race -covermode=atomic"
        output: "coverage.out"
    docker:
      # docker build/push related, default to false
      enable: False
      # dockerhub image repo
      repo: golang-demo

```

## How to Reproduce

Steps to reproduce the bug:

1. create a new git repo via dtm
2. delete git repo from github
3. run again the apply -f
4. See the error 

## Expected Behavior
expect is should can create success

## Screenshots

If applicable, add screenshots to help explain the issue.

## Additional Context

Add any other helpful context about the issue here.

daniel-hutao:
@JaneLiuL I went into your `repository list` and didn't find a project named `test-demo`. Please confirm that the project has been successfully created. 

It is also possible that the delete operation is performed twice.

JaneLiuL:
i think you have some mistakes.
please apply first, then delete repo on github,
then apply again, will fail at 404.
[...]

PR #266, opened 2022-03-04
bug: github actions and repo scaffolding apply error after deleting repo

IronCore864:
# Summary

See issue #263

### New Behavior

```bash
tiexin@MBA ~/work/stream $ ./dtm-darwin-arm64 apply
2022-03-04 10:41:27 ℹ [INFO]  Apply started.
2022-03-04 10:41:27 ℹ [INFO]  Using dir <.devstream> to store plugins.
2022-03-04 10:41:31 ℹ [INFO]  Tool < go-webapp-repo > state found but it seems the tool isn't created, will be created.
2022-03-04 10:41:31 ℹ [INFO]  Tool < golang-demo-actions > state found but it seems the tool isn't created, will be created.
Continue? [y/n]
Enter a value (Default is n): n
```

### DIFF
--- a/internal/pkg/plugin/githubactions/golang/read.go
+++ b/internal/pkg/plugin/githubactions/golang/read.go
@@ -23,13 +23,12 @@ func Read(options map[string]interface{}) (map[string]interface{}, error) {
 	}
 
 	path, err := gitHubClient.GetWorkflowPath()
-	if path == "" && err == nil {
-		// file not found
-		return nil, nil
-	}
-
 	if err != nil {
 		return nil, err
+	}
+	if path == "" {
+		// file not found
+		return nil, nil
 	}
 
 	log.Debugf("Language is: %s.", ga.GetLanguage(opt.Language))
--- a/internal/pkg/plugin/githubactions/golang/read.go
+++ b/internal/pkg/plugin/githubactions/golang/read.go
@@ -23,6 +23,11 @@ func Read(options map[string]interface{}) (map[string]interface{}, error) {
 	}
 
 	path, err := gitHubClient.GetWorkflowPath()
+	if path == "" && err == nil {
+		// file not found
+		return nil, nil
+	}
+
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


internal/pkg/plugin/githubactions/golang/read.go
```diff
@@ -23,6 +23,11 @@ func Read(options map[string]interface{}) (map[string]interface{}, error) {
 	}
 
 	path, err := gitHubClient.GetWorkflowPath()
+	if path == "" && err == nil {
```
daniel-hutao, line 26:
```go
if err != nil{
 return nil, err
}
if path == ""{
 return nil, nil
}
```

I think it's better.

pkg/util/github/repo.go
```diff
@@ -23,11 +22,16 @@ func (c *Client) CreateRepo() error {
 }
 
 func (c *Client) Delete() error {
-	_, err := c.Client.Repositories.Delete(
+	response, err := c.Client.Repositories.Delete(
```
daniel-hutao, line 25: Our code below uses `resp` to represent the `response`, maybe it can be consistent here, and use `resp` anywhere. Conversely, the following should also use the name `response`

internal/pkg/plugin/githubactions/python/read.go
```diff
@@ -23,7 +23,12 @@ func Read(options map[string]interface{}) (map[string]interface{}, error) {
 	}
 
 	path, err := ghClient.GetWorkflowPath()
-	if err != nil {
+	if path == "" && err == nil {
+		// file not found
+		return nil, nil
+	}
+
+  if err != nil {
```
lfbdev, line 31: Format can be adjusted

lfbdev (approved):

pkg/util/github/repo.go
IronCore864: I actually prefer the full word here, and there are 3 reasons:

1. response is more widely used than resp (339M vs 62M according to GitHub)
2. Although in this context, "resp" isn't ambiguous, in other contexts, "resp" can mean other things, like respTabId -> responsive tab id. We should reduce ambiguity to make our code as easy to read as possible.
3. Compared to "response", "resp" isn't a lot shorter. Using the abbreviation doesn't save a whole lot of line width. If it's a tradition (like using "i" for loop counter) or can significantly reduce the line length (e.g., i18n vs internationalization, I think we can use the abbreviation)

So, I suggest in the future we can gradually replace "resp" to response.

daniel-hutao: 👍

daniel-hutao (approved):

Issue #264, opened 2022-03-04
GitHub Actions Plugin Fails When Docker Section Doesn't Exist


IronCore864:
## Describe the Bug

When docker isn't disabled explicitly, `dtm apply` fails.

## How to Reproduce

Steps to reproduce the bug:

`config.yaml`:

```yaml
---
tools:
- name: golang-demo-actions
  plugin:
    kind: githubactions-golang
    version: 0.2.0
  options:
    owner: IronCore864
    repo: go-webapp-devstream-demo
    language:
      name: go
      version: "1.17"
    branch: main
    build:
      enable: True
    test:
      enable: True
      coverage:
        enable: True
```

Note that there isn't docker enabled true because I want to use default value false.

Then `dtm apply`, got some logs:

```bash
...
2022-03-04 09:18:26 ℹ [INFO]  Processing: golang-demo-actions -> Create ...
2022-03-04 09:18:26 ✖ [ERROR]  Param error: docker is empty.
2022-03-04 09:18:26 ✖ [ERROR]  Plugin golang-demo-actions Create failed.
2022-03-04 09:18:26 ℹ [INFO]  golang-demo-actions-Create -> incorrect params
...

```

## Expected Behavior

When the docker section is omitted from the config, there shouldn't be an error; we can use the default value for it, which is enabled = False.

## Screenshots

If applicable, add screenshots to help explain the issue.

<img width="307" alt="Screenshot 2022-03-04 at 09 28 09" src="https://user-images.githubusercontent.com/19277614/156681638-26b3d25c-ac1c-49fd-9abd-620ccc27a66c.png">

<img width="834" alt="Screenshot 2022-03-04 at 09 28 30" src="https://user-images.githubusercontent.com/19277614/156681643-30ac27cf-77cf-472b-9bcc-e9abd76d02bd.png">

## Workaround

For users who don't want to enable Docker, temporarily, we can add:

```yaml
    docker:
      enable: False
```

to the "options" section of the config.

PR #268, opened 2022-03-04
bug: GitHub Actions Plugin fails when Docker section doesn't exist

lfbdev:
# Summary

when docker section config is empty, dtm will raise errors
image

## Description
resolution
1. remove docker section validation
2. add if condition
3. when docker section is empty in render template, let it be false

### Related Issues

#264

### DIFF
--- a/internal/pkg/plugin/githubactions/golang/options.go
+++ b/internal/pkg/plugin/githubactions/golang/options.go
@@ -53,14 +53,5 @@ func validate(param *Options) []error {
 		}
 	}
 
-	if param.Docker == nil {
-		retErrors = append(retErrors, fmt.Errorf("docker is empty"))
-	}
-	if errs := param.Docker.Validate(); len(errs) != 0 {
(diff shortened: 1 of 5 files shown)


internal/pkg/plugin/githubactions/golang/options.go
```diff
@@ -53,14 +53,5 @@ func validate(param *Options) []error {
 		}
 	}
 
-	if param.Docker == nil {
```
daniel-hutao, line 56: if param.Docker == nil {
 	return retErrors
}

if errs := param.Docker.Validate(); len(errs) != 0 {
 	for _, e := range errs {
 		retErrors = append(retErrors, fmt.Errorf("docker is invalid: %s", e))
 	}
 }
return retErrors

daniel-hutao (changes requested):

IronCore864 (approved):

daniel-hutao (approved):

PR #277, opened 2022-03-05
refactor: using `util/github` in `trellogithub`

daniel-hutao:

# Summary

There are many functions in the `trellogithub` package that are duplicated in `util/github`, so we can delete these functions and directly refer to `util/github`.

But there are still some functions that are not implemented in `util/github`, so part of the functions is not easy to refactor for now. Later we need to spend some time for letting the `trellogithub` without any line of code directly use the `"github.com/google/go-github/v42/github"`.


lfbdev (approved):
/lgtm

IronCore864 (approved):

Issue #278, opened 2022-03-06
Feat: add "dtm delete --force" for deleting everything defined in the config, regardless of the state


lfbdev:
## Description

for deleting everything defined in the config, regardless of the state

add an option for delete cmd : dtm delete --force

PR #279, opened 2022-03-06
feat: add 'dtm delete --force' for deleting everything

lfbdev:
# Summary

1. to delete everything defined in config
2. ignore state in devstream.state file
3.  --force, no shorthand

### Related Issues

#278

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ Similarly, to delete without confirmation:
 To delete everything defined in the config, regardless of the state: 
 
 ```bash
-./dtm delete -r -f YOUR_CONFIG_FILE.yaml
+./dtm delete --force -f YOUR_CONFIG_FILE.yaml
 ```
 
 To verify, run:
--- a/README.md
+++ b/README.md
@@ -123,6 +123,11 @@ Similarly, to delete without confirmation:
 ```bash
 ./dtm -y delete -f YOUR_CONFIG_FILE.yaml
 ```
+To delete everything defined in the config, regardless of the state: 
+
+```bash
+./dtm delete -r -f YOUR_CONFIG_FILE.yaml
+```
 
 To verify, run:
 
--- a/cmd/devstream/main.go
+++ b/cmd/devstream/main.go
@@ -16,7 +16,7 @@ var (
 	pluginDir        string
 	continueDirectly bool
 	isDebug          bool
-	isForce          bool
+	isForceDelete    bool
 
 	rootCMD = &cobra.Command{
 		Use:   "dtm",
@@ -44,7 +44,6 @@ func init() {
 	rootCMD.PersistentFlags().StringVarP(&pluginDir, "plugin-dir", "p", pluginengine.DefaultPluginDir, "plugins directory")
 	rootCMD.PersistentFlags().BoolVarP(&continueDirectly, "yes", "y", false, "apply/delete directly without confirmation")
 	rootCMD.PersistentFlags().BoolVarP(&isDebug, "debug", "d", false, "debug level log")
-	rootCMD.PersistentFlags().BoolVarP(&isForce, "force", "r", false, "force to delete from config")
 	rootCMD.AddCommand(versionCMD)
 	rootCMD.AddCommand(initCMD)
 	rootCMD.AddCommand(applyCMD)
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -123,6 +123,11 @@ Similarly, to delete without confirmation:
 ```bash
 ./dtm -y delete -f YOUR_CONFIG_FILE.yaml
 ```
+To delete everything defined in the config, regardless of the state: 
+
+```bash
+./dtm delete -r -f YOUR_CONFIG_FILE.yaml
```
IronCore864, line 129: I think we need a better parameter here.

"-f" is taken by "--from-file" but "-r" doesn't relate to "force" that much.

What do you think?

I suggest --force, no short form, because "-f" is taken, and because force delete is dangerous, so it doesn't hurt to let the user type more to make sure this is really what they want.

cmd/devstream/main.go
```diff
@@ -43,6 +44,7 @@ func init() {
 	rootCMD.PersistentFlags().StringVarP(&pluginDir, "plugin-dir", "p", pluginengine.DefaultPluginDir, "plugins directory")
 	rootCMD.PersistentFlags().BoolVarP(&continueDirectly, "yes", "y", false, "apply/delete directly without confirmation")
 	rootCMD.PersistentFlags().BoolVarP(&isDebug, "debug", "d", false, "debug level log")
+	rootCMD.PersistentFlags().BoolVarP(&isForce, "force", "r", false, "force to delete from config")
```
IronCore864, line 47: "isForce" should not belong to "rootCMD". It should belong to "delete" subcommand only, because we won't have "dtm apply --force".

README.md
daniel-hutao, line 129: I think `--force` is better than `-r`

internal/pkg/pluginengine/change.go
```diff
@@ -40,27 +40,31 @@ const (
 // GetChangesForApply takes "State Manager" & "Config" then do some calculate and return a Plan.
 // All actions should be execute is included in this Plan.changes.
 func GetChangesForApply(smgr statemanager.Manager, cfg *configloader.Config) ([]*Change, error) {
-	return getChanges(smgr, cfg, CommandApply)
+	return getChanges(smgr, cfg, CommandApply, false)
```
IronCore864, line 43: if "isForce" needs to be passed to "getChanges", then the name isn't good enough. Because according to the logic of "getChanges", it can get changes for both delete, and apply. What does "isForce" mean exactly? Is apply force? Or is delete force? Or both?

We could have a better naming, like isForceDelete, or simply forceDelete. Or, we can refactor getChanges so that it doesn't depends on forceDelete.

internal/pkg/pluginengine/change_helper.go
```diff
@@ -136,3 +136,19 @@ func changesForDelete(smgr statemanager.Manager, cfg *configloader.Config) []*Ch
...
 	return changes
 }
+
+// changesForForceDelete  for force delete from config, ignore state file
+func changesForForceDelete(smgr statemanager.Manager, cfg *configloader.Config) []*Change {
+	changes := make([]*Change, 0)
+	tmpStates := smgr.GetStatesMap().DeepCopy()
+
+	log.Debug("tools: ", cfg.Tools)
+	//
```
IronCore864, line 146: redundant comment

IronCore864 (changes requested):

internal/pkg/pluginengine/change.go
```diff
@@ -40,27 +40,31 @@ const (
...
-
+	log.Debug("isForce:", isForce)
 	// calculate changes from config and state
 	var changes []*Change
 	var err error
 	if commandType == CommandApply {
 		changes, err = changesForApply(smgr, cfg)
 	} else if commandType == CommandDelete {
-		changes = changesForDelete(smgr, cfg)
+		if isForce {
```
daniel-hutao, line 63: `changes = changesForDelete(smgr, cfg, isForce)`

I think it's better to write it like the above; of course this is just my personal opinion and it doesn't have to be write this way. you can decide for yourself.

daniel-hutao (changes requested):
all.

internal/pkg/pluginengine/change.go
lfbdev: I think in this format,  if isForce=true, 'change'  will be generated twice

IronCore864 (approved):

daniel-hutao (approved):

Issue #284, opened 2022-03-07
Proposal: integrate openldap plugin


imxw:
## Description

Most of the DevOps tools used by small and medium-sized companies are open source and have their own user authentication system. If each platform maintains a set of user data, the maintenance workload will increase greatly, so we basically choose a unified authentication system, and [OpenLDAP](https://www.openldap.org/) is a good choice, and it is free and open source. As a DevOps toolchain manager, it is necessary for DevStream to integrate the tool, and I will provide some of my best practices.

daniel-hutao:
Thanks for your contribution.

/assign @imxw

PR #292, opened 2022-03-08
feat: add openldap plugin

imxw:
# Summary

- add new plugin OpenLDAP which can achieve unified identity authentication
- add related documents
## Key Points

- [x] Documentation (new or existing) is updated.

### Related Issues

#284


lfbdev (commented):
build_linux_amd64.sh  should be updated.

Add  such as "go build -buildmode=plugin -trimpath -gcflags="all=-N -l" -o .devstream/openldap-${GOOS}-${GOARCH}_${VERSION}.so ..." into the file

daniel-hutao (approved):

daniel-hutao:
LGTM!!!

imxw:
> build_linux_amd64.sh should be updated.
> 
> Add such as "go build -buildmode=plugin -trimpath -gcflags="all=-N -l" -o .devstream/openldap-${GOOS}-${GOARCH}_${VERSION}.so ..." into the file

Yes, I have modified the file.

imxw:
> > build_linux_amd64.sh should be updated.
> > Add such as "go build -buildmode=plugin -trimpath -gcflags="all=-N -l" -o .devstream/openldap-${GOOS}-${GOARCH}_${VERSION}.so ..." into the file
> 
> Yes, I have modified the file.

Got it, I only changed the Makefile

lfbdev (approved):

IronCore864:
Very good, thanks a lot! @imxw 

We are going to do some manual local tests before merging. Will keep you posted.

IronCore864:
Hello @imxw, great job! This will be our first community plugin and it comes much earlier than we expected. Kudos!

I have 3 suggestions after testing:

1. In the doc, in the example config, maybe add `replicaCount: 1` to the `values_yaml` section, because most people would like to test it locally before deploying it elsewhere, and it's highly likely that their local environment is a one-node mini K8s cluster like `minikube` or `kind`.
2. For the same reason, in the example config, in the `values_yaml` section, change `persistence.enabled` to `false`.
3. The `values_yaml` section contains _a lot of_ default values from this helm chart, for example, `LDAP_LOG_LEVEL`, etc. When installing the chart, the values_yaml will be used on top of the default values from this chart, so it's not necessary to copy the default values from this chart. If the `values_yaml` section is too long, the doc isn't easy to read.

Please rethink what's the best way to put a short and easy-to-ready example in the doc with only necessary values.

IronCore864:
One more small issue on the doc: please pay attention to "[title case](https://en.wikipedia.org/wiki/Title_case)".

E.g., instead of "Description of key fields in values_yaml", "Description of Key Fields in `values_yaml`" is much better. If you are unsure about it, try the [title case converter](https://titlecaseconverter.com/).

@imxw

IronCore864 (changes requested):
No more comments; see the conversation for details.

imxw:
> Hello @imxw, great job! This will be our first community plugin and it comes much earlier than we expected. Kudos!
> 
> I have 3 suggestions after testing:
> 
> 1. In the doc, in the example config, maybe add `replicaCount: 1` to the `values_yaml` section, because most people would like to test it locally before deploying it elsewhere, and it's highly likely that their local environment is a one-node mini K8s cluster like `minikube` or `kind`.
> 2. For the same reason, in the example config, in the `values_yaml` section, change `persistence.enabled` to `false`.
> 3. The `values_yaml` section contains _a lot of_ default values from this helm chart, for example, `LDAP_LOG_LEVEL`, etc. When installing the chart, the values_yaml will be used on top of the default values from this chart, so it's not necessary to copy the default values from this chart. If the `values_yaml` section is too long, the doc isn't easy to read.
> 
> Please rethink what's the best way to put a short and easy-to-ready example in the doc with only necessary values.

Thank you for your suggestion, I will make some changes to the document

IronCore864 (approved):

IronCore864:
Closes #284

PR #293, opened 2022-03-08
feat: plugin dependency topological sort

IronCore864:
# Summary

## Key Points

- [x] Documentation (new or existing) is updated. TODO

## Description

Topologically sorting the tools before `dtm apply/delete/destroy`.

### Current Behavior

Currently, for `dtm apply`, the changes are executed in the order that the tools are defined in the config. For `dtm delete`, the order is reversed (as a hack so that no dependency issue happens).

### New Behavior

The tools which have no dependency will be executed first; then the tools whose dependency have been resolved in the last execution; so on and so forth.

### Tests

Normal apply/delete without any dependency - same behavior as before.

Dependency doesn't exist in the config:

```bash
2022-03-10 17:18:59 ℹ [INFO]  Apply started.
2022-03-10 17:18:59 ✖ [ERROR]  Config validation failed: tool go-webapp-repo-2's dependency abc.def doesn't exist in the config.
2022-03-10 17:18:59 ✖ [ERROR]  Apply failed => failed to load the config file.
```

Dependency loop:

```bash
2022-03-10 17:21:16 ℹ [INFO]  Apply started.
2022-03-10 17:21:16 ℹ [INFO]  Using dir <.devstream> to store plugins.
2022-03-10 17:21:16 ✖ [ERROR]  Apply failed => dependency loop detected in the config.
```


internal/pkg/pluginengine/change_helper.go
```diff
@@ -59,49 +59,60 @@ func drifted(a, b map[string]interface{}) bool {
...
 
-	// 1, create a temporary state map used to store unprocessed tools.
+	// 1. create a temporary state map used to store unprocessed tools.
 	tmpStatesMap := smgr.GetStatesMap().DeepCopy()
 
-	// 2, for each tool in the config, generate changes.
-	for _, tool := range cfg.Tools {
-		state := smgr.GetState(statemanager.StateKeyGenerateFunc(&tool))
+	// 2. handle dependency and sort the tools in the config
+	var tools [][]configloader.Tool
```
daniel-hutao, line 66: The tools with type `[][]configloader.Tool` and the tools with type `[]configloader.Tool` in `topologicalSort(tools []configloader.Tool)` confuse me a bit. Same name but different type here. Maybe there's a better way to name it

internal/pkg/pluginengine/change_helper.go
```diff
@@ -59,49 +59,60 @@ func drifted(a, b map[string]interface{}) bool {
...
+	// the elements in tools are sorted "batches"
+	// and each element
```
daniel-hutao, line 68: This sentence does not seem to be very long, why is it divided into 2 lines?

daniel-hutao (commented):
all above

internal/pkg/pluginengine/change_helper.go
```diff
@@ -59,49 +59,60 @@ func drifted(a, b map[string]interface{}) bool {
...
+						// tool exists in the state, but resource doesn't exist, Create
+						description := fmt.Sprintf("Tool < %s > state found but it seems the tool isn't created, will be created.", tool.Name)
+						changes = append(changes, generateCreateAction(&tool, description))
+					} else if drifted(resource, state.Resource) {
+						// resource drifted from state, Update
+						description := fmt.Sprintf("Tool < %s > drifted from the state, will be updated.", tool.Name)
+						changes = append(changes, generateUpdateAction(&tool, description))
+					} else {
+						// resource is the same as the state, do nothing
+						log.Debugf("Tool < %s > is the same as the state, do nothing.", tool.Name)
```
lfbdev, line 108: does nothing？

internal/pkg/pluginengine/topological_sort.go
```diff
@@ -0,0 +1,82 @@
...
+			}
+
+			// if there isn't any dependency: it's the "start" of the graph
+			// we can put it into the first batch
+			if len(tool.DependsOn) == 0 {
+				log.Debugf("TOOL %s.%s dependency already solved\n", tool.Name, tool.Plugin.Kind)
+				batch = append(batch, tool)
+			} else {
+				dependency_resolved := true
+				for _, dep := range tool.DependsOn {
```
lfbdev, line 46: Too many nesting levels,  should a new function be abstracted?

internal/pkg/pluginengine/change_helper.go
IronCore864: I think this is fine. It doesn't sound official enough, but it means something like "the tool is the same as the state, so we do nothing here (so nothing is done here)".

internal/pkg/pluginengine/topological_sort.go
IronCore864: Very good point!

daniel-hutao (approved):

lfbdev (approved):

PR #300, opened 2022-03-09
doc: fix flux case problem

Underwood-M:
Signed-off-by: Underwood-M <dongfanghuifan@gmail.com>

# Summary

doc: fix flux case problem


IronCore864:
Hi @Underwood-M, thanks for pointing this out! Your attention to detail is extraordinary:)

I also did some research, it seems "Flux" should be uppercase.

I'm not so sure about "FluxCD" or "Flux CD" (with or without the space). According to Google, it seems to be "Flux CD" (see attached screenshot), but I didn't find it elsewhere. What do you think?

Screenshot 2022-03-10 at 16 34 53

Underwood-M:
image
image

daniel-hutao (approved):

IronCore864 (approved):

PR #313, opened 2022-03-14
feat: separate plug-in for create Trello board

lfbdev:
# Summary

1.  separate plug-in for create Trello board, add outputs

## Description

separate plug-in for create Trello board, add outputs

### Related Issues

#314 
### New Behavior

image

### DIFF
--- a/cmd/trello/main.go
+++ b/cmd/trello/main.go
@@ -1,0 +1,39 @@
+package main
+
+import (
+	"github.com/merico-dev/stream/internal/pkg/plugin/trello"
+	"github.com/merico-dev/stream/pkg/util/log"
+)
+
+// NAME is the name of this DevStream plugin.
+const NAME = "trello"
+
+// Plugin is the type used by DevStream core. It's a string.
+type Plugin string
+
+// Create implements the creation of trello board.
+func (p Plugin) Create(options map[string]interface{}) (map[string]interface{}, error) {
+	return trello.Create(options)
+}
+
+// Update implements the creation of trello board.
+func (p Plugin) Update(options map[string]interface{}) (map[string]interface{}, error) {
+	return trello.Update(options)
+}
+
+// Read implements the healthy check of trello board.
+func (p Plugin) Read(options map[string]interface{}) (map[string]interface{}, error) {
+	return trello.Read(options)
+}
+
+// Delete implements the creation of trello board.
+func (p Plugin) Delete(options map[string]interface{}) (bool, error) {
+	return trello.Delete(options)
+}
+
+// DevStreamPlugin is the exported variable used by the DevStream core.
+var DevStreamPlugin Plugin
+
+func main() {
+	log.Infof("%T: %s is a plugin for DevStream. Use it with DevStream.\n", NAME, DevStreamPlugin)
+}
--- a/docs/plugins/trello_plugin.md
+++ b/docs/plugins/trello_plugin.md
@@ -1,0 +1,77 @@
+## 1 `trello` Plugin
+
+This plugin creates a new Trello board and lists.
+
+## 2 Usage:
+
+_This plugin depends on the following two environment variables:_
+
+- TRELLO_API_KEY
+- TRELLO_TOKEN
+
+Set the values accordingly before using this plugin.
+
+## 3 Tips:
+_Trello board description is managed by DevStream, please don't modify it._
+
+To create a Trello API key and token, see [here](https://trello.com/app-key).
+
+```yaml
+tools:
+  - name: default
+    # plugin profile
+    plugin:
+      # kind of this plugin
+      kind: trello
+      # version of the plugin
+      version: 0.2.0
+    # options for the plugin, checkout the version from the GitHub releases
+    options:
+      # the repo's owner (if kanbanBoardName is empty, use owner/repo as the boardname)
+      owner: lfbdev
+      # the repo name  (if kanbanBoardName is empty, use owner/repo as the boardname)
+      repo: golang-demo
+      # the Tello board name 
+      kanbanBoardName: kanban-name
+```
+
+## 3. Use Together with the `trello-github-integ` Plugin
+
+This plugin can be used together with the `trello-github-integ` plugin (see document [here](./trello-github-integ_plugin.md).)
+
+See the example below:
+
+```yaml
+---
+tools:
+  - name: default
+    plugin:
+      kind: trello
+      version: 0.2.0
+    options:
+      owner: lfbdev
+      repo: golang-demo
+      kanbanBoardName: kanban-name
+  - name: default_trello_github
+    plugin:
+      kind: trello-github-integ
+      version: 0.2.0
+    dependsOn: ["default.trello"]
+    options:
+      owner: lfbdev
+      repo: golang-demo
+      api:
+        name: trello
+        boardId: ${{ default.trello.outputs.bid }}
+        todoListId: ${{ default.trello.outputs.todoid }}
+        doingListId: ${{ default.trello.outputs.doingid }}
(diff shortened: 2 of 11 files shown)


internal/pkg/plugin/trello/trello_test.go
```diff
@@ -0,0 +1,9 @@
+package trello
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+)
+
+var _ = Describe("Trello", func() {
```
IronCore864, line 7: Remove this file for now; we can add it later.

internal/pkg/plugin/trello/helper.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/mitchellh/mapstructure"
+	
+	"github.com/merico-dev/stream/pkg/util/log"
+	"github.com/merico-dev/stream/pkg/util/trello"
+)
+
+func buildState(options *Options, ti *TrelloItemId) map[string]interface{} {
+	res := make(map[string]interface{})
+	output := make(map[string]interface{})
```
IronCore864, line 14: The definition can be moved closer to the first usage (above line 20) so that it's easier to read.

cmd/trello/main.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/merico-dev/stream/pkg/util/log"
+)
+
+// NAME is the name of this DevStream plugin.
+const NAME = "trello"
+
+// Plugin is the type used by DevStream core. It's a string.
+type Plugin string
+
+// Create implements the installation of trello board.
```
IronCore864, line 14: "installation": creation might be better because Trello is a web service; we don't really "install" anything. Also, all usages of "Trello" should be the same as their official name (uppercase T).

cmd/trello/main.go
```diff
@@ -0,0 +1,39 @@
...
+func (p Plugin) Create(options map[string]interface{}) (map[string]interface{}, error) {
+	return trello.Create(options)
+}
+
+// Update implements the installation of trello board.
```
IronCore864, line 19: Wrong comment.

cmd/trello/main.go
```diff
@@ -0,0 +1,39 @@
...
+func (p Plugin) Update(options map[string]interface{}) (map[string]interface{}, error) {
+	return trello.Update(options)
+}
+
+// Read implements the healthy check of trello board.
+func (p Plugin) Read(options map[string]interface{}) (map[string]interface{}, error) {
+	return trello.Read(options)
+}
+
+// Delete implements the installation of trello board.
```
IronCore864, line 29: Wrong comment.

internal/pkg/plugin/trello/create.go
```diff
@@ -0,0 +1,28 @@
+package trello
+
+import (
+	"github.com/merico-dev/stream/pkg/util/log"
+)
+
+// Create sets up trello.
```
IronCore864, line 7: "sets up" isn't quite self-explanatory. It's better to put the exact behavior, such as, "it creates a Trello board and some columns", etc.

internal/pkg/plugin/trello/create.go
```diff
@@ -0,0 +1,28 @@
...
+func Create(options map[string]interface{}) (map[string]interface{}, error) {
+
```
IronCore864, line 9: This empty new line isn't quite necessary.

internal/pkg/plugin/trello/delete.go
```diff
@@ -0,0 +1,7 @@
+package trello
+
+// Delete remove trello.
+func Delete(options map[string]interface{}) (bool, error) {
+
+	return true, nil
```
IronCore864, line 6: Why is there no implementation? Is this by design?

internal/pkg/plugin/trello/trello.go
```diff
@@ -0,0 +1,62 @@
...
+
+type TrelloItemId struct {
+	boardId     string
+	todoListId  string
+	doingListId string
+	doneListId  string
+}
+
+// CreateTrelloItems create board/lists, and set secret by ids
+func CreateTrelloItems(options *Options) (*TrelloItemId, error) {
```
IronCore864, line 25: As far as I can see from the code, this piece of code only creates a "board"; no "lists" are created. So, the comment might not be very accurate. Also, the name of the function is "CreateTrelloItems". What does "Item" mean here? Does "Item" mean both "board" and "lists"? Is this an official usage? I personally suggest a better function name.

internal/pkg/plugin/trello/update.go
```diff
@@ -0,0 +1,28 @@
+package trello
+
+import (
+	"github.com/merico-dev/stream/pkg/util/log"
+)
+
+// Update remove and set up trello workflows.
```
IronCore864, line 7: Wrong comment? I don't think this plugin has anything to do with workflow any more.

internal/pkg/plugin/trello/update.go
```diff
@@ -0,0 +1,28 @@
...
+
+	if err := validateOptions(opt); err != nil {
+		return nil, err
+	}
+
+	trelloIds, err := CreateTrelloItems(opt)
+	if err != nil {
+		return nil, err
+	}
+	log.Success("Creating trello board succeeded.")
```
IronCore864, line 25: I think in the core, we have logging for the success of each plugin, right? So, this might be redundant information? Please run some tests to make sure the logging output makes sense to a user (i.e., order, no redundant info, etc.)

daniel-hutao:
I have seen a lot of suggestions from @IronCore864 , so I will wait until these `conversation` are resolved before starting the review to avoid falling into "duplicate problems". @lfbdev Please feel free to call me when you are ready for review again.

internal/pkg/plugin/trello/helper.go
```diff
@@ -0,0 +1,60 @@
...
+	res["boardId"] = ti.boardId
+	res["todoListId"] = ti.todoListId
+	res["doingListId"] = ti.doingListId
+	res["doneListId"] = ti.doneListId
+
+	output["bid"] = ti.boardId
```
IronCore864, line 20: "bid" is too simple to show its real meaning. How about boardId/boardID? Pay attention to upper/lowercase. todoid->todoId, doingid->doingId, etc.

IronCore864:
Please also update the document.

internal/pkg/plugin/trello/delete.go
lfbdev: In the previous discussion, it was said that this board cannot be deleted because someone might be using it

internal/pkg/plugin/trello/delete.go
```diff
@@ -0,0 +1,7 @@
+package trello
+
+// Delete remove trello.
+func Delete(options map[string]interface{}) (bool, error) {
+
```
daniel-hutao, line 5: Should we add some log here? Do nothing when delete is a bit confusing.

daniel-hutao (commented):
see the comment above.

docs/plugins/trello_plugin.md
```diff
@@ -0,0 +1,77 @@
...
+        todoListId: ${{ default.trello.outputs.todoid }}
+        doingListId: ${{ default.trello.outputs.doingid }}
+        doneListId: ${{ default.trello.outputs.doneid }}
+      branch: main
+```
+
+In the example above:
+
+- We put `default.trello` as dependency by using the `dependsOn` keyword.
+- We used `default.trello`'s output as input for the `default_trello_github` plugin.
```
lfbdev: use

IronCore864 (approved):

daniel-hutao (approved):

PR #323, opened 2022-03-15
feat: add plugin outputs reference

lfbdev:
# Summary

Use format: ${{ TOOL_NAME.TOOL_KIND.outputs.var}} to ref the outputs of other plugins

## Description

### Related Issues

#324 

### Screenshots

image

### DIFF
--- a/docs/plugins/trello-github-integ_plugin.md
+++ b/docs/plugins/trello-github-integ_plugin.md
@@ -1,23 +1,10 @@
 ## 1 `trello-github-integ` Plugin
 
-This plugin creates a new Trello board and integrate it with your GitHub repo.
+This plugin creates a new github actions file(trello-github-integration) and upload to your GitHub repo.
 
 ## 2 Usage:
 
-_This plugin depends on the following two environment variables:_
-
-- TRELLO_API_KEY
-- TRELLO_TOKEN
-
-Set the values accordingly before using this plugin.
-
-If you don't know how to create these two tokens, check out:
-- [Generate a Trello API key and token](https://docs.servicenow.com/bundle/quebec-it-asset-management/page/product/software-asset-management2/task/generate-trello-apikey-token.html)
-
-## 3 Tips:
-_Trello board description is managed by DevStream, please don't modify it._
-
-To create a Trello API key and token, see [here](https://docs.servicenow.com/bundle/quebec-it-asset-management/page/product/software-asset-management2/task/generate-trello-apikey-token.html).
+_This plugin depends on the plugin `trello`:_
 
 ```yaml
 tools:
@@ -37,13 +24,51 @@ tools:
     owner: YOUR_GITHUB_USERNAME
     # the repo where you'd like to setup GitHub Actions; please change the value below.
     repo: YOUR_REPO_NAME
-    # integration tool name
-    api:
-      name: trello
-      # name of the Trello kanban board
-      kanbanBoardName: kanban-name
+    # reference from dependency
+    boardId: ${{ default.trello.outputs.boardId }}
+    todoListId: ${{ default.trello.outputs.todoListId }}
+    doingListId: ${{ default.trello.outputs.doingListId }}
+    doneListId: ${{ default.trello.outputs.doneListId }}
     # main branch of the repo (to which branch the plugin will submit the workflows)
     branch: main
 ```
 
-Currently, all the parameters in the example above are mandatory.
+## 3. Use Together with the `trello` Plugin
+
+This plugin can be used together with the `trello` plugin (see document [here](./trello_plugin.md).)
+
+See the example below:
+
+```yaml
+---
+tools:
+  - name: default
+    plugin:
+      kind: trello
+      version: 0.2.0
+    options:
+      owner: lfbdev
+      repo: golang-demo
+      kanbanBoardName: kanban-name
+  - name: default_trello_github
+    plugin:
+      kind: trello-github-integ
+      version: 0.2.0
+    dependsOn: ["default.trello"]
+    options:
+      owner: lfbdev
+      repo: golang-demo
+      boardId: ${{ default.trello.outputs.bid }}
+      todoListId: ${{ default.trello.outputs.todoid }}
+      doingListId: ${{ default.trello.outputs.doingid }}
+      doneListId: ${{ default.trello.outputs.doneid }}
+      branch: main
+```
+
+In the example above:
+
+- We put `default.trello` as dependency by using the `dependsOn` keyword.
+- We use `default.trello`'s output as input for the `default_trello_github` plugin.
+
+Pay attention to the `${{ xxx }}` part in the example. `${{ TOOL_NAME.TOOL_KIND.outputs.var}}` is the syntax for using an output.
+
--- a/docs/plugins/trello-github-integ_plugin.md
+++ b/docs/plugins/trello-github-integ_plugin.md
@@ -1,6 +1,6 @@
 ## 1 `trello-github-integ` Plugin
 
-This plugin creates a new GitHub Actions file(trello-github-integration) and upload to your GitHub repo.
+This plugin creates a new GitHub Actions workflow(trello-github-integration) and uploads it to your GitHub repo.
 
 ## 2 Usage:
 
@@ -17,7 +17,7 @@ tools:
     # checkout the version from the GitHub releases
     version: 0.2.0
   # optional; if specified, dtm will make sure the dependency is applied first before handling this tool.
-  dependsOn: [ "TOOL1_NAME.TOOL1_KIND", "TOOL2_NAME.TOOL2_KIND" ]
+  dependsOn: [ "TRELLO_INSTANCE_NAME.trello" ]
   # options for the plugin
   options:
     # the repo's owner. It should be case-sensitive here; strictly use your GitHub user name; please change the value below.
@@ -25,10 +25,10 @@ tools:
     # the repo where you'd like to setup GitHub Actions; please change the value below.
     repo: YOUR_REPO_NAME
     # reference parameters come from dependency, their usage will be explained later
-    boardId: ${{ default.trello.outputs.boardId }}
-    todoListId: ${{ default.trello.outputs.todoListId }}
-    doingListId: ${{ default.trello.outputs.doingListId }}
-    doneListId: ${{ default.trello.outputs.doneListId }}
+    boardId: ${{ TRELLO_INSTANCE_NAME.trello.outputs.boardId }}
+    todoListId: ${{ TRELLO_INSTANCE_NAME.trello.outputs.todoListId }}
+    doingListId: ${{ TRELLO_INSTANCE_NAME.trello.outputs.doingListId }}
+    doneListId: ${{ TRELLO_INSTANCE_NAME.trello.outputs.doneListId }}
     # main branch of the repo (to which branch the plugin will submit the workflows)
     branch: main
 ```
@@ -67,8 +67,8 @@ tools:
 
 In the example above:
 
-- We put `default.trello` as dependency by using the `dependsOn` keyword.
-- We use `default.trello`'s output as input for the `default_trello_github` plugin.
+- We put `trello_init_demo.trello` as a dependency by using the `dependsOn` keyword.
+- We use `trello_init_demo.trello`'s output as input for the `trello_github_integ_demo` plugin.
 
 Pay attention to the `${{ xxx }}` part in the example. `${{ TOOL_NAME.TOOL_KIND.outputs.var}}` is the syntax for using an output.
 
--- a/docs/plugins/trello-github-integ_plugin.md
+++ b/docs/plugins/trello-github-integ_plugin.md
@@ -1,6 +1,6 @@
 ## 1 `trello-github-integ` Plugin
 
-This plugin creates a new github actions file(trello-github-integration) and upload to your GitHub repo.
+This plugin creates a new GitHub Actions file(trello-github-integration) and upload to your GitHub repo.
 
 ## 2 Usage:
 
@@ -24,7 +24,7 @@ tools:
     owner: YOUR_GITHUB_USERNAME
     # the repo where you'd like to setup GitHub Actions; please change the value below.
     repo: YOUR_REPO_NAME
-    # reference from dependency
+    # reference parameters come from dependency, their usage will be explained later
     boardId: ${{ default.trello.outputs.boardId }}
     todoListId: ${{ default.trello.outputs.todoListId }}
     doingListId: ${{ default.trello.outputs.doingListId }}
@@ -42,7 +42,7 @@ See the example below:
 ```yaml
 ---
 tools:
-  - name: default
+  - name: trello_init_demo
     plugin:
       kind: trello
       version: 0.2.0
@@ -50,18 +50,18 @@ tools:
       owner: lfbdev
       repo: golang-demo
       kanbanBoardName: kanban-name
-  - name: default_trello_github
+  - name: trello_github_integ_demo
     plugin:
       kind: trello-github-integ
       version: 0.2.0
-    dependsOn: ["default.trello"]
+    dependsOn: ["trello_init_demo.trello"]
     options:
       owner: lfbdev
       repo: golang-demo
-      boardId: ${{ default.trello.outputs.bid }}
-      todoListId: ${{ default.trello.outputs.todoid }}
-      doingListId: ${{ default.trello.outputs.doingid }}
-      doneListId: ${{ default.trello.outputs.doneid }}
+      boardId: ${{ trello_init_demo.trello.outputs.boardId }}
+      todoListId: ${{ trello_init_demo.trello.outputs.todoListId }}
+      doingListId: ${{ trello_init_demo.trello.outputs.doingListId }}
+      doneListId: ${{ trello_init_demo.trello.outputs.doneListId }}
       branch: main
 ```
 
--- a/internal/pkg/plugin/trello/trello.go
+++ b/internal/pkg/plugin/trello/trello.go
@@ -28,6 +28,25 @@ func CreateTrelloBoard(options *Options) (*TrelloItemId, error) {
 		return nil, err
 	}
 
+	exist, err := c.CheckBoardExists(options.Owner, options.Repo, options.KanbanBoardName)
+	if err != nil {
+		return nil, err
+	}
+
+	if exist {
+		log.Infof("Board already exists, owner: %s, repo: %s, kanbanName: %s.", options.Owner, options.Repo, options.KanbanBoardName)
+		listIds, err := c.GetBoardIdAndListId(options.Owner, options.Repo, options.KanbanBoardName)
+		if err != nil {
+			return nil, err
+		}
+		return &TrelloItemId{
+			boardId:     fmt.Sprint(listIds["boardId"]),
+			todoListId:  fmt.Sprint(listIds["todoListId"]),
+			doingListId: fmt.Sprint(listIds["doingListId"]),
+			doneListId:  fmt.Sprint(listIds["doneListId"]),
+		}, nil
+	}
+
 	board, err := c.CreateBoard(options.KanbanBoardName, options.Owner, options.Repo)
 	if err != nil {
 		return nil, err
--- a/internal/pkg/plugin/trellogithub/create.go
+++ b/internal/pkg/plugin/trellogithub/create.go
@@ -11,25 +11,17 @@ func Create(options map[string]interface{}) (map[string]interface{}, error) {
 		return nil, err
 	}
 
-	api := tg.GetApi()
-	log.Infof("API is: %s.", api.Name)
-	ws := defaultWorkflows.GetWorkflowByNameVersionTypeString(api.Name)
-
-	for _, w := range ws {
-		if err := tg.renderTemplate(w); err != nil {
-			return nil, err
-		}
-		if err := tg.client.AddWorkflow(w, tg.options.Branch); err != nil {
-			return nil, err
-		}
+	if err := tg.client.AddWorkflow(trelloWorkflow, tg.options.Branch); err != nil {
+		return nil, err
 	}
+
 	log.Success("Adding workflow file succeeded.")
 
 	trelloIds := &TrelloItemId{
-		boardId:     tg.GetApi().BoardId,
-		todoListId:  tg.GetApi().todoListId,
-		doingListId: tg.GetApi().doingListId,
-		doneListId:  tg.GetApi().doneListId,
+		boardId:     tg.options.BoardId,
+		todoListId:  tg.options.todoListId,
+		doingListId: tg.options.doingListId,
+		doneListId:  tg.options.doneListId,
 	}
 
 	if err := tg.AddTrelloIdSecret(trelloIds); err != nil {
--- a/internal/pkg/plugin/trellogithub/create.go
+++ b/internal/pkg/plugin/trellogithub/create.go
@@ -24,11 +24,14 @@ func Create(options map[string]interface{}) (map[string]interface{}, error) {
 		}
 	}
 	log.Success("Adding workflow file succeeded.")
-	trelloIds, err := tg.CreateTrelloItems()
-	if err != nil {
-		return nil, err
+
+	trelloIds := &TrelloItemId{
+		boardId:     tg.GetApi().BoardId,
+		todoListId:  tg.GetApi().todoListId,
+		doingListId: tg.GetApi().doingListId,
+		doneListId:  tg.GetApi().doneListId,
 	}
-	log.Success("Creating trello board succeeded.")
+
 	if err := tg.AddTrelloIdSecret(trelloIds); err != nil {
 		return nil, err
 	}
--- a/internal/pkg/plugin/trellogithub/create.go
+++ b/internal/pkg/plugin/trellogithub/create.go
@@ -17,18 +17,18 @@ func Create(options map[string]interface{}) (map[string]interface{}, error) {
 
 	log.Success("Adding workflow file succeeded.")
 
-	trelloIds := &TrelloItemId{
+	trelloItemId := &TrelloItemId{
 		boardId:     tg.options.BoardId,
 		todoListId:  tg.options.todoListId,
 		doingListId: tg.options.doingListId,
 		doneListId:  tg.options.doneListId,
 	}
 
-	if err := tg.AddTrelloIdSecret(trelloIds); err != nil {
+	if err := tg.AddTrelloIdSecret(trelloItemId); err != nil {
 		return nil, err
 	}
 
 	log.Success("Adding secret keys for trello succeeded.")
 
-	return buildState(tg, trelloIds), nil
+	return buildState(tg), nil
 }
--- a/internal/pkg/plugin/trellogithub/helper.go
+++ b/internal/pkg/plugin/trellogithub/helper.go
@@ -9,7 +9,6 @@ import (
 	"github.com/mitchellh/mapstructure"
 
 	"github.com/merico-dev/stream/pkg/util/github"
-	"github.com/merico-dev/stream/pkg/util/trello"
 )
 
 // renderTemplate render the github actions template with config.yaml
@@ -37,27 +36,18 @@ func (tg *TrelloGithub) renderTemplate(workflow *github.Workflow) error {
 func buildState(tg *TrelloGithub, ti *TrelloItemId) map[string]interface{} {
 	res := make(map[string]interface{})
 	res["workflowDir"] = fmt.Sprintf("/repos/%s/%s/contents/.github/workflows", tg.options.Owner, tg.options.Repo)
-	res["boardId"] = ti.boardId
-	res["todoListId"] = ti.todoListId
-	res["doingListId"] = ti.doingListId
-	res["doneListId"] = ti.doneListId
 	return res
 }
 
 func (tg *TrelloGithub) buildReadState(api *Api) (map[string]interface{}, error) {
-	c, err := trello.NewClient()
-	if err != nil {
-		return nil, err
-	}
-	listIds, err := c.GetBoardIdAndListId(tg.options.Owner, tg.options.Repo, api.KanbanBoardName)
-	if err != nil {
-		return nil, err
-	}
+
+	listIds := make(map[string]interface{})
 
 	path, err := tg.GetWorkflowPath()
 	if err != nil {
 		return nil, err
 	}
 	listIds["workflowDir"] = path
+
 	return listIds, nil
 }
--- a/internal/pkg/plugin/trellogithub/helper.go
+++ b/internal/pkg/plugin/trellogithub/helper.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 )
 
-func buildState(tg *TrelloGithub, ti *TrelloItemId) map[string]interface{} {
+func buildState(tg *TrelloGithub) map[string]interface{} {
 	res := make(map[string]interface{})
 	res["workflowDir"] = fmt.Sprintf("/repos/%s/%s/contents/.github/workflows", tg.options.Owner, tg.options.Repo)
 	return res
--- a/internal/pkg/plugin/trellogithub/helper.go
+++ b/internal/pkg/plugin/trellogithub/helper.go
@@ -1,37 +1,8 @@
 package trellogithub
 
 import (
-	"bytes"
 	"fmt"
-
-	"text/template"
-
-	"github.com/mitchellh/mapstructure"
-
-	"github.com/merico-dev/stream/pkg/util/github"
 )
-
-// renderTemplate render the github actions template with config.yaml
-func (tg *TrelloGithub) renderTemplate(workflow *github.Workflow) error {
-	var jobs Jobs
-	err := mapstructure.Decode(tg.options.Jobs, &jobs)
-	if err != nil {
-		return err
-	}
-	//if use default {{.}}, it will confict (github actions vars also use them)
-	t, err := template.New("trellogithub").Delims("[[", "]]").Parse(workflow.WorkflowContent)
-	if err != nil {
-		return err
-	}
-
-	var buff bytes.Buffer
-	err = t.Execute(&buff, jobs)
-	if err != nil {
-		return err
-	}
-	workflow.WorkflowContent = buff.String()
-	return nil
-}
 
 func buildState(tg *TrelloGithub, ti *TrelloItemId) map[string]interface{} {
 	res := make(map[string]interface{})
@@ -39,8 +10,7 @@ func buildState(tg *TrelloGithub, ti *TrelloItemId) map[string]interface{} {
 	return res
 }
 
-func (tg *TrelloGithub) buildReadState(api *Api) (map[string]interface{}, error) {
-
+func (tg *TrelloGithub) buildReadState() (map[string]interface{}, error) {
 	listIds := make(map[string]interface{})
 
 	path, err := tg.GetWorkflowPath()
--- a/internal/pkg/plugin/trellogithub/update.go
+++ b/internal/pkg/plugin/trellogithub/update.go
@@ -30,15 +30,17 @@ func Update(options map[string]interface{}) (map[string]interface{}, error) {
 		}
 	}
 	log.Success("Adding workflow file succeeded.")
-	trelloIds, err := tg.CreateTrelloItems()
-	if err != nil {
-		return nil, err
+
+	trelloIds := &TrelloItemId{
+		boardId:     tg.GetApi().BoardId,
+		todoListId:  tg.GetApi().todoListId,
+		doingListId: tg.GetApi().doingListId,
+		doneListId:  tg.GetApi().doneListId,
 	}
-	log.Success("Creating trello board succeeded.")
+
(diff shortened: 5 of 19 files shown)


internal/pkg/plugin/trellogithub/create.go
```diff
@@ -24,11 +24,14 @@ func Create(options map[string]interface{}) (map[string]interface{}, error) {
 		}
 	}
 	log.Success("Adding workflow file succeeded.")
-	trelloIds, err := tg.CreateTrelloItems()
-	if err != nil {
-		return nil, err
+
+	trelloIds := &TrelloItemId{
+		boardId:     tg.GetApi().BoardId,
```
IronCore864, line 29: I think "GetApi" is a little ambiguous. What does it get exactly? An API client?

Based on the code, I think it returns a struct, which contains some information, which are board ID, list IDs, etc. They have nothing to do with "API". I think we need a better naming here.

IronCore864: Also, where does the boardID, list IDs come from?

It seems they are part of the "options", but the documentation isn't updated. The doc should include an example, letting users know what parameters are required in the "options" section.

An example of how to refer to trello outputs is also required. Refer to my previous pull request for an example: https://github.com/merico-dev/stream/pull/316/files.

internal/pkg/plugin/trellogithub/helper.go
```diff
@@ -37,27 +36,18 @@ func (tg *TrelloGithub) renderTemplate(workflow *github.Workflow) error {
...
 func (tg *TrelloGithub) buildReadState(api *Api) (map[string]interface{}, error) {
-	c, err := trello.NewClient()
-	if err != nil {
-		return nil, err
-	}
-	listIds, err := c.GetBoardIdAndListId(tg.options.Owner, tg.options.Repo, api.KanbanBoardName)
-	if err != nil {
-		return nil, err
-	}
+
```
IronCore864, line 43: This seems like an unnecessary empty new line.

internal/pkg/plugin/trellogithub/update.go
```diff
@@ -30,15 +30,17 @@ func Update(options map[string]interface{}) (map[string]interface{}, error) {
 		}
 	}
 	log.Success("Adding workflow file succeeded.")
-	trelloIds, err := tg.CreateTrelloItems()
-	if err != nil {
-		return nil, err
+
+	trelloIds := &TrelloItemId{
+		boardId:     tg.GetApi().BoardId,
+		todoListId:  tg.GetApi().todoListId,
```
IronCore864, line 36: Same issue with "create".

internal/pkg/plugin/trellogithub/workflow.go
```diff
@@ -34,8 +34,11 @@ type Options struct {
 }
 
 type Api struct {
-	Name            string
-	KanbanBoardName string
+	Name        string
```
IronCore864, line 37: The name of the struct says "API" but according to the members of the struct, they don't seem like API related.

internal/pkg/pluginengine/change.go
```diff
@@ -169,3 +182,46 @@ func handleResult(smgr statemanager.Manager, change *Change) error {
 	log.Successf("Plugin %s %s done.", change.Tool.Name, change.ActionName)
 	return nil
 }
+
+// FillInputParams fill inputs from state
+func FillInputParams(smgr statemanager.Manager, options map[string]interface{}) error {
+	//traverse options
+	for key, value := range options {
+		log.Debug("Key: ", key, ", Value: ", value)
+		// judge wheter the value is string
```
IronCore864, line 191: wheter -> whether, string -> a string

internal/pkg/pluginengine/change.go
```diff
@@ -169,3 +182,46 @@ func handleResult(smgr statemanager.Manager, change *Change) error {
 	log.Successf("Plugin %s %s done.", change.Tool.Name, change.ActionName)
 	return nil
 }
+
+// FillInputParams fill inputs from state
+func FillInputParams(smgr statemanager.Manager, options map[string]interface{}) error {
+	//traverse options
+	for key, value := range options {
+		log.Debug("Key: ", key, ", Value: ", value)
```
IronCore864, line 190: Missing period (".") at the end of the log sentense.

internal/pkg/pluginengine/change.go
```diff
@@ -169,3 +182,46 @@ func handleResult(smgr statemanager.Manager, change *Change) error {
 	log.Successf("Plugin %s %s done.", change.Tool.Name, change.ActionName)
 	return nil
 }
+
+// FillInputParams fill inputs from state
+func FillInputParams(smgr statemanager.Manager, options map[string]interface{}) error {
+	//traverse options
```
IronCore864, line 188: //traverse -> // traverse

missing space.

internal/pkg/pluginengine/change.go
```diff
@@ -169,3 +182,46 @@ func handleResult(smgr statemanager.Manager, change *Change) error {
...
+	for key, value := range options {
+		log.Debug("Key: ", key, ", Value: ", value)
+		// judge wheter the value is string
+		if inst, ok := value.(string); ok {
+			// judge wheter the format is ${{}}
```
IronCore864, line 193: wheter -> whether;

The format isn't exactly ${{}}, but rather, ${{xxx}}. So the comment here isn't quite accurate.

internal/pkg/pluginengine/change.go
```diff
@@ -169,3 +182,46 @@ func handleResult(smgr statemanager.Manager, change *Change) error {
...
+			if strings.HasPrefix(inst, REF_PREFIX) && strings.HasSuffix(inst, REF_SUFFIX) {
+				ref := strings.TrimSpace(strings.TrimSuffix(strings.TrimPrefix(inst, REF_PREFIX), REF_SUFFIX))
+				log.Debug("Ref inputs: ", ref)
+				refParam := strings.Split(ref, ".")
+				if len(refParam) <= 3 {
```
IronCore864, line 198: What is "3"? This is a magic number. The logic of handling the output reference doesn't need to be placed here.

internal/pkg/pluginengine/change.go
```diff
@@ -169,3 +182,46 @@ func handleResult(smgr statemanager.Manager, change *Change) error {
...
+					return errors.New("ref input format is not correct: " + ref)
+				}
+
+				state := smgr.GetState(genStateKey(refParam))
+				outputs := state.Resource["outputs"]
```
IronCore864, line 203: Line 202, 203: maybe we can have a "GetOutputs" function for the state manager, so that we don't have to do two things here. It makes the code more precise.

internal/pkg/pluginengine/change.go
```diff
@@ -169,3 +182,46 @@ func handleResult(smgr statemanager.Manager, change *Change) error {
...
+				log.Debug("Ref outputs: ", outputs)
+				if outs, ok := outputs.(map[string]interface{}); ok {
+					log.Debug("Ref outs: ", outs)
+					log.Debug("Ref param: ", refParam[3])
+					options[key] = outs[refParam[3]]
```
IronCore864, line 208: Magic number. Plus, what if refParam[3] is not in the "outs"?

internal/pkg/pluginengine/change.go
```diff
@@ -169,3 +182,46 @@ func handleResult(smgr statemanager.Manager, change *Change) error {
...
+					if value == nil {
+						return errors.New("ref input value is null: " + refParam[3])
+					}
+				}
+			}
+		}
+		// judge wheter the format is map[string]interface{}, if true, then recursion
```
IronCore864, line 215: I do not understand here. I thought we have already replaced outputs to their values in the previous loop. What is the purpose of this recursion?

internal/pkg/pluginengine/change.go
```diff
@@ -169,3 +182,46 @@ func handleResult(smgr statemanager.Manager, change *Change) error {
...
+		if _, ok := value.(map[string]interface{}); ok {
+			if err := FillInputParams(smgr, value.(map[string]interface{})); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func genStateKey(refParam []string) statemanager.StateKey {
```
IronCore864, line 225: I recall vaguely that we have a similar function from the statemanager.
Maybe this can be moved to the state manager as well.

internal/pkg/pluginengine/change.go
```diff
@@ -96,6 +101,14 @@ func execute(smgr statemanager.Manager, changes []*Change) map[string]error {
 		var err error
 		var returnValue map[string]interface{}
 
+		log.Info("tool raw changes: ", c.Tool.Options)
+		// fill ref inputs
+		err = FillInputParams(smgr, c.Tool.Options)
```
IronCore864, line 106: I think this is probably too late. Maybe we can move the logic into configloader. Update the outputs in-place immediately after the conf is loaded. If there are errors in the outputs, for example, if the key doesn't exist in the outputs at all, an error can be thrown out earlier. We do not have to wait till "change" (apply/delete).

IronCore864 (changes requested):

IronCore864:
I also suggest that we add some unit tests to cover this change. We can add many irregular test cases so that abnormal user input won't break our program.

internal/pkg/pluginengine/change.go
lfbdev: This was done before, but there are some issues.

Because the reference ${{}}  depends on the result of the previous plugin.

internal/pkg/pluginengine/change.go
lfbdev: ${{ TOOL_NAME.TOOL_KIND.outputs.var}},  when it is splitted to be []string,  element 3 is the var

internal/pkg/pluginengine/change.go
lfbdev: ${{ TOOL_NAME.TOOL_KIND.outputs.var}},  param lenth cannot <=3

internal/pkg/pluginengine/change.go
lfbdev: Because the structure in options may be as follows, not necessarily the first layer can find ${{xxx}}, maybe the second layer, maybe the third layer.

eg.
1. in first layer:
  options:
      doingListId: ${{ default.trello.outputs.doingListId }}
      doneListId: ${{ default.trello.outputs.doneListId }}

2 in second layer:
  options:
    app:
      name: good_name
      doingListId: ${{ default.trello.outputs.doingListId }}
      doneListId: ${{ default.trello.outputs.doneListId }}

internal/pkg/plugin/trellogithub/create.go
lfbdev: 1. will modify 
2. boardID, list IDs come from ${{xxx}}.  will update the doc

docs/plugins/trello-github-integ_plugin.md
```diff
@@ -1,23 +1,10 @@
 ## 1 `trello-github-integ` Plugin
 
-This plugin creates a new Trello board and integrate it with your GitHub repo.
+This plugin creates a new github actions file(trello-github-integration) and upload to your GitHub repo.
```
IronCore864, line 3: github actions -> [GitHub Actions](https://github.com/features/actions)

We should use the official name.

docs/plugins/trello-github-integ_plugin.md
```diff
@@ -37,13 +24,51 @@ tools:
...
+tools:
+  - name: default
+    plugin:
+      kind: trello
+      version: 0.2.0
+    options:
+      owner: lfbdev
+      repo: golang-demo
+      kanbanBoardName: kanban-name
+  - name: default_trello_github
```
IronCore864, line 53: Let's give it a more meaningful name than default. Same for Trello plugin.

docs/plugins/trello-github-integ_plugin.md
```diff
@@ -37,13 +24,51 @@ tools:
...
+    plugin:
+      kind: trello-github-integ
+      version: 0.2.0
+    dependsOn: ["default.trello"]
+    options:
+      owner: lfbdev
+      repo: golang-demo
+      boardId: ${{ default.trello.outputs.bid }}
```
IronCore864, line 61: I don't think we have "bid" any more. Please keep it consistent with the code.

docs/plugins/trello-github-integ_plugin.md
```diff
@@ -37,13 +24,51 @@ tools:
     owner: YOUR_GITHUB_USERNAME
     # the repo where you'd like to setup GitHub Actions; please change the value below.
     repo: YOUR_REPO_NAME
-    # integration tool name
-    api:
-      name: trello
-      # name of the Trello kanban board
-      kanbanBoardName: kanban-name
+    # reference from dependency
+    boardId: ${{ default.trello.outputs.boardId }}
```
IronCore864, line 28: At this point, the user hasn't read the "outputs" part yet, so they might be confused by this syntax. Maybe we can:
- either add a comment to point it out that this is referring to outputs, which is explained later;
- or use some placeholders

internal/pkg/pluginengine/change.go
```diff
@@ -96,6 +101,14 @@ func execute(smgr statemanager.Manager, changes []*Change) map[string]error {
 		var err error
```
IronCore864, line 101: I think we do not need to define "err" here; we can use "err := xxx" below.

internal/pkg/pluginengine/change.go
```diff
@@ -169,3 +182,41 @@ func handleResult(smgr statemanager.Manager, change *Change) error {
 	log.Successf("Plugin %s(%s) %s done.", change.Tool.Name, change.Tool.Plugin.Kind, change.ActionName)
 	return nil
 }
+
+// FillInputParams fill inputs from state
+func FillInputParams(smgr statemanager.Manager, options map[string]interface{}) error {
```
IronCore864, line 187: This function doesn't need to be "exported", because it's only being used within the same package.

internal/pkg/pluginengine/change.go
IronCore864: This is very hard for others to understand what exactly `if var < 3` means.

How about: 

`if isValidOutputReference(str)` ?

internal/pkg/pluginengine/change.go
IronCore864: I do not understand. Could you give a concrete example with a full config.yaml?

Correct me if I'm wrong:

If B depends on A and B uses A's outputs, A will be executed first. When we execute B, we can already get all the outputs in one pass. Even if C depends on B and uses B's outputs, by the time C gets executed, B already finished, so we can fill C's input with one pass.

internal/pkg/pluginengine/change.go
IronCore864: I might also suggest a better name. "Fill input params" doesn't describe what it does exactly. handle outputs, fill outputs? Not very good examples, please think about it more.

internal/pkg/statemanager/manager.go
```diff
@@ -28,6 +28,7 @@ type Manager interface {
 	AddState(key StateKey, state State) error
 	UpdateState(key StateKey, state State) error
 	DeleteState(key StateKey) error
+	GetOutputs(refParam []string) interface{}
```
IronCore864, line 31: Maybe we should keep the parameter here as StateKey instead of []string, to make it more consistent?

IronCore864: Maybe return two parameters, adding "error" as well. In case outputs isn't in the state.

internal/pkg/statemanager/manager.go
```diff
@@ -106,3 +107,8 @@ func (m *manager) DeleteState(key StateKey) error {
 	m.statesMap.Delete(key)
 	return m.Write(m.GetStatesMap().Format())
 }
+
+func (m *manager) GetOutputs(refParam []string) interface{} {
```
IronCore864, line 111: This is an important function that should be covered by unit tests.

internal/pkg/pluginengine/change.go
IronCore864: This is an important function that should be covered by unit tests.

IronCore864:
FYI: it seems the latest lint failed. Please fix.

docs/plugins/trello-github-integ_plugin.md
```diff
@@ -1,23 +1,10 @@
 ## 1 `trello-github-integ` Plugin
 
-This plugin creates a new Trello board and integrate it with your GitHub repo.
+This plugin creates a new GitHub Actions file(trello-github-integration) and upload to your GitHub repo.
```
IronCore864, line 3: Please use Grammarly on all updated documents.

internal/pkg/pluginengine/change.go
```diff
@@ -96,6 +101,14 @@ func execute(smgr statemanager.Manager, changes []*Change) map[string]error {
 		var err error
 		var returnValue map[string]interface{}
 
+		log.Info("tool raw changes: ", c.Tool.Options)
```
IronCore864, line 104: Let's keep "logs" (not errors, to be specific) as a "full sentence", which means:

- first letter as uppercase
- period at the end

Same for logs below.

internal/pkg/pluginengine/change.go
```diff
@@ -169,3 +182,47 @@ func handleResult(smgr statemanager.Manager, change *Change) error {
 	log.Successf("Plugin %s(%s) %s done.", change.Tool.Name, change.Tool.Plugin.Kind, change.ActionName)
 	return nil
 }
+
+// FillInputParams fill inputs from state
```
IronCore864, line 186: Comment/function name mismatch.

internal/pkg/pluginengine/change.go
```diff
@@ -169,3 +182,47 @@ func handleResult(smgr statemanager.Manager, change *Change) error {
...
+func fillRefValueWithOutputs(smgr statemanager.Manager, options map[string]interface{}) error {
+	// traverse options
```
IronCore864, line 188: I think this line of comment is redundant. `for ... range ...` is already quite self-explanatory.

internal/pkg/pluginengine/change.go
```diff
@@ -169,3 +182,47 @@ func handleResult(smgr statemanager.Manager, change *Change) error {
...
+	for key, value := range options {
+		log.Debugf("Key: %s,  Value: %s.", key, value)
+		// judge whether the value is a string
+		if inst, ok := value.(string); ok {
+			// judge whether the format is ${{xxx}}
+			if strings.HasPrefix(inst, REF_PREFIX) && strings.HasSuffix(inst, REF_SUFFIX) {
```
IronCore864, line 194: A suggestion: something like `if isValidOutputsReference(inst)` would make this line shorter, easier to understand, and the comment might not be needed in this case.

internal/pkg/pluginengine/change.go
```diff
@@ -169,3 +182,47 @@ func handleResult(smgr statemanager.Manager, change *Change) error {
...
+				ref := strings.TrimSpace(strings.TrimSuffix(strings.TrimPrefix(inst, REF_PREFIX), REF_SUFFIX))
```
IronCore864, line 195: This train of trims makes it hard to read. Maybe we can create a function and give it a meaningful name.

internal/pkg/pluginengine/change.go
```diff
@@ -169,3 +182,47 @@ func handleResult(smgr statemanager.Manager, change *Change) error {
...
+				log.Debug("Ref inputs: ", ref)
+				refParam := strings.Split(ref, ".")
+				if len(refParam) <= 3 {
+					return errors.New("ref input format is not correct: " + ref)
```
IronCore864, line 199: Errors, unlike logs, don't have to be a full sentence, which means they can be a sentence fragment with missing elements (like verbs). "incorrect output reference" is more than enough.

internal/pkg/statemanager/state.go
```diff
@@ -67,3 +67,7 @@ type StateKey string
 func StateKeyGenerateFunc(t *configloader.Tool) StateKey {
 	return StateKey(fmt.Sprintf("%s_%s", t.Name, t.Plugin.Kind))
 }
+
+func GenStateKey(refParam []string) StateKey {
```
IronCore864, line 71: "refParam" might be understandable (although not the best name) in the context of the pluginengine, but in the statemanager, it means nothing. The parameter's name can be improved.

internal/pkg/pluginengine/change.go
IronCore864: Please also check log lines in other files.

internal/pkg/pluginengine/fillRefValueWithOutputs_test.go
```diff
@@ -0,0 +1,223 @@
...
+		Plugin:   configloader.Plugin{Kind: "trello-github-integ", Version: "0.2.0"},
+		Options:  map[string]interface{}{"a": "value"},
+		Resource: map[string]interface{}{"a": "value"},
+	}
+
+	keystateReferee := statemanager.StateKey("default_trello")
+	stateReferee := statemanager.State{
+		Name:     "default",
+		Plugin:   configloader.Plugin{Kind: "trello", Version: "0.2.0"},
+		Options:  map[string]interface{}{"owner": "lfbdev", "repo": "golang-demo", "kanbanBoardName": "kanban-name"},
```
IronCore864, line 27: Writing them in multiple lines instead of one single line makes it much easier to read. Code is for humans to read, not for machines to execute.

internal/pkg/pluginengine/fillRefValueWithOutputs_test.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/stretchr/testify/assert"
+
+	"github.com/merico-dev/stream/internal/pkg/configloader"
+	"github.com/merico-dev/stream/internal/pkg/statemanager"
+)
+
+// TestOutputsInState normal test, if the ref value is correct
+func TestOutputsInState(t *testing.T) {
+	keyReferrer := statemanager.StateKey("default-trello-github_trello-github-integ")
+	stateReferrer := statemanager.State{
```
IronCore864, line 16: It seems this part is duplicated code across all test cases. Similar for the defer part. Maybe we can utilize setup/teardown mechanism for this

internal/pkg/pluginengine/fillRefValueWithOutputs_test.go
```diff
@@ -0,0 +1,223 @@
...
+		Name:     "default-trello-github",
+		Plugin:   configloader.Plugin{Kind: "trello-github-integ", Version: "0.2.0"},
+		Options:  map[string]interface{}{"a": "value"},
+		Resource: map[string]interface{}{"a": "value"},
+	}
+
+	keystateReferee := statemanager.StateKey("default_trello")
```
IronCore864, line 23: A referee is a judge in a sports game. ReferredState/Reference might be a better name.

internal/pkg/pluginengine/fillRefValueWithOutputs_test.go
```diff
@@ -0,0 +1,223 @@
...
+			log.Errorf("failed to delete state %s.", keyReferrer)
+		}
+		if err := smgr.DeleteState(keystateReferee); err != nil {
+			log.Errorf("failed to delete state %s.", keystateReferee)
+		}
+	}()
+
+	options := map[string]interface{}{"owner": "lfbdev", "repo": "golang-demo", "branch": "main", "app": map[string]interface{}{"boardId": "${{ default.trello.outputs.boardId }}", "todoListId": "${{ default.trello.outputs.todoListId }}"}}
+
+	expectResult := map[string]interface{}{"owner": "lfbdev", "repo": "golang-demo", "branch": "main", "app": map[string]interface{}{"boardId": "123", "todoListId": "456"}}
```
IronCore864, line 92: Question: what is "app" here? I don't understand. What is the difference between this test case and the previous one?

internal/pkg/pluginengine/fillRefValueWithOutputs_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+	options := map[string]interface{}{"owner": "lfbdev", "repo": "golang-demo", "branch": "main", "boardId": "${{ default.trello.outputs.boardId }}", "todoListId": "${{ default.trello.outputs.todoListId }}"}
+
+	err = fillRefValueWithOutputs(smgr, options)
+
+	assert.Equal(t, "cannot find outputs from state: default", err.Error())
+}
+
+// TestWrongRefFormat exception test, if the outputs do not contain ref
+func TestWrongRefFormat(t *testing.T) {
```
IronCore864, line 142: I wouldn't call this test "wrong format", because the outputs reference format is indeed correct. I think it's only because the referred key is not in the outputs.

internal/pkg/pluginengine/fillRefValueWithOutputs_test.go
lfbdev: previous is "options- ${{ }}",  this case is  "options- app- ${{ }}",   to test refs  in different levels

docs/plugins/trello-github-integ_plugin.md
```diff
@@ -1,23 +1,10 @@
...
-Set the values accordingly before using this plugin.
-
-If you don't know how to create these two tokens, check out:
-- [Generate a Trello API key and token](https://docs.servicenow.com/bundle/quebec-it-asset-management/page/product/software-asset-management2/task/generate-trello-apikey-token.html)
-
-## 3 Tips:
-_Trello board description is managed by DevStream, please don't modify it._
-
-To create a Trello API key and token, see [here](https://docs.servicenow.com/bundle/quebec-it-asset-management/page/product/software-asset-management2/task/generate-trello-apikey-token.html).
+_This plugin depends on the plugin `trello`:_
```
daniel-hutao, line 7: LIne 20: `dependsOn: [ "TOOL1_NAME.TOOL1_KIND", "TOOL2_NAME.TOOL2_KIND" ]`

Maybe this line could be updated to:

`dependsOn: [ "TRELLO_INSTANCE_NAME.trello" ]`

docs/plugins/trello-github-integ_plugin.md
```diff
@@ -37,13 +24,51 @@ tools:
...
+## 3. Use Together with the `trello` Plugin
+
+This plugin can be used together with the `trello` plugin (see document [here](./trello_plugin.md).)
+
+See the example below:
+
+```yaml
+---
+tools:
+  - name: trello_init_demo
```
daniel-hutao, line 45: The name is `trello_init_demo` here, but I saw the `default.trello` at line 70/71.

So the name is `default` or `trello_init_demo`?

docs/plugins/trello-github-integ_plugin.md
```diff
@@ -37,13 +24,51 @@ tools:
...
     # the repo where you'd like to setup GitHub Actions; please change the value below.
     repo: YOUR_REPO_NAME
-    # integration tool name
-    api:
-      name: trello
-      # name of the Trello kanban board
-      kanbanBoardName: kanban-name
+    # reference parameters come from dependency, their usage will be explained later
+    boardId: ${{ default.trello.outputs.boardId }}
+    todoListId: ${{ default.trello.outputs.todoListId }}
```
daniel-hutao, line 29: Why the name is `default` here?
I think it may be `TRELLO_INSTANCE_NAME`.

internal/pkg/plugin/trello/trello.go
```diff
@@ -28,6 +28,25 @@ func CreateTrelloBoard(options *Options) (*TrelloItemId, error) {
 		return nil, err
 	}
 
+	exist, err := c.CheckBoardExists(options.Owner, options.Repo, options.KanbanBoardName)
+	if err != nil {
+		return nil, err
+	}
+
+	if exist {
+		log.Successf("Board already exists, owner: %s, repo: %s, kanbanName: %s.", options.Owner, options.Repo, options.KanbanBoardName)
```
daniel-hutao, line 37: If the board has already existed, we do nothing here, so the `log.Successf()` isn't very suitable. I think `Info or Debug` log level is better.

daniel-hutao: For the same reason, we can move the `log.Successf()` after the board/list is created successfully below.

internal/pkg/plugin/trellogithub/create.go
```diff
@@ -11,24 +11,19 @@ func Create(options map[string]interface{}) (map[string]interface{}, error) {
...
+	if err := tg.client.AddWorkflow(trelloWorkflow, tg.options.Branch); err != nil {
+		return nil, err
 	}
+
 	log.Success("Adding workflow file succeeded.")
-	trelloIds, err := tg.CreateTrelloItems()
-	if err != nil {
-		return nil, err
+
+	trelloIds := &TrelloItemId{
```
daniel-hutao, line 20: The instance with struct `TrelloItemId` should be `trelloItemId` or something else but not `xxxs`, if the plural is used, it seems like the instance with `[]TrelloItemId` or `map[xxx]TrelloItemId`

internal/pkg/plugin/trellogithub/helper.go
```diff
@@ -1,62 +1,23 @@
...
-	var buff bytes.Buffer
-	err = t.Execute(&buff, jobs)
-	if err != nil {
-		return err
-	}
-	workflow.WorkflowContent = buff.String()
-	return nil
-}
-
 func buildState(tg *TrelloGithub, ti *TrelloItemId) map[string]interface{} {
```
daniel-hutao, line 7: The `ti` param is not used since the `ti.xxx` has been deleted. so it should be removed.

internal/pkg/plugin/trellogithub/helper.go
```diff
@@ -1,62 +1,23 @@
...
-func (tg *TrelloGithub) buildReadState(api *Api) (map[string]interface{}, error) {
-	c, err := trello.NewClient()
-	if err != nil {
-		return nil, err
-	}
-	listIds, err := c.GetBoardIdAndListId(tg.options.Owner, tg.options.Repo, api.KanbanBoardName)
-	if err != nil {
-		return nil, err
-	}
+func (tg *TrelloGithub) buildReadState() (map[string]interface{}, error) {
```
daniel-hutao, line 13: The functions `buildState(tg *TrelloGithub, ti *TrelloItemId) map[string]interface{}` and `func (tg *TrelloGithub) buildReadState() (map[string]interface{}, error)` are so similar, I kind of don't see why two functions are needed here to do what seems to be the same thing.

internal/pkg/plugin/trellogithub/update.go
```diff
@@ -11,34 +11,28 @@ func Update(options map[string]interface{}) (map[string]interface{}, error) {
...
+
+	err = tg.client.AddWorkflow(trelloWorkflow, tg.options.Branch)
 	if err != nil {
 		return nil, err
 	}
-	log.Success("Creating trello board succeeded.")
+
+	log.Success("Adding workflow file succeeded.")
+
+	trelloIds := &TrelloItemId{
```
daniel-hutao, line 26: Same as above. I think the name `trelloIds` is unsuitable here.

internal/pkg/pluginengine/change.go
```diff
@@ -96,6 +101,14 @@ func execute(smgr statemanager.Manager, changes []*Change) map[string]error {
 		var err error
 		var returnValue map[string]interface{}
 
+		log.Infof("Tool's raw changes are: %s.", c.Tool.Options)
```
daniel-hutao, line 104: The `Options` here is a map, so what will happen with printing it as a string(%s)?

internal/pkg/pluginengine/change.go
```diff
@@ -169,3 +182,56 @@ func handleResult(smgr statemanager.Manager, change *Change) error {
 	log.Successf("Plugin %s(%s) %s done.", change.Tool.Name, change.Tool.Plugin.Kind, change.ActionName)
 	return nil
 }
+
+// fillRefValueWithOutputs fill inputs from state
+func fillRefValueWithOutputs(smgr statemanager.Manager, options map[string]interface{}) error {
```
daniel-hutao, line 187: Too many `if` in this function, it's hard to read.

internal/pkg/pluginengine/change_helper.go
```diff
@@ -166,3 +169,30 @@ func changesForForceDelete(smgr statemanager.Manager, cfg *configloader.Config)
 	}
 	return changes
 }
+
+func renderRefByDependency(tool *configloader.Tool, tools []configloader.Tool, smgr statemanager.Manager) error {
```
daniel-hutao, line 173: Too many if in this function, it's hard to read.

internal/pkg/pluginengine/fillRefValueWithOutputs_test.go
```diff
@@ -0,0 +1,221 @@
+package pluginengine
```
daniel-hutao, line 1: We should use the `ginkgo` framework for the ut.

internal/pkg/pluginengine/getOutputs_test.go
```diff
@@ -0,0 +1,89 @@
+package pluginengine
```
daniel-hutao, line 1: We should use the ginkgo framework for the ut.

internal/pkg/statemanager/manager.go
```diff
@@ -106,3 +107,14 @@ func (m *manager) DeleteState(key StateKey) error {
 	m.statesMap.Delete(key)
 	return m.Write(m.GetStatesMap().Format())
 }
+
+func (m *manager) GetOutputs(key StateKey) (interface{}, error) {
+	state := m.GetState(key)
+	if state == nil {
+		return nil, fmt.Errorf("cannot find state by key: %s", key)
+	}
+	if value, ok := state.Resource["outputs"]; ok {
```
daniel-hutao, line 116: if value, ok := state.Resource["outputs"]; !ok {
return nil, fmt.Errorf("cannot find outputs from state: %s", state.Name)
}
return value, nil

pkg/util/trello/trello.go
```diff
@@ -78,8 +78,23 @@ func (c *Client) GetBoardIdAndListId(owner, repo, kanbanBoardName string) (map[s
...
+		if checkTargetBoard(owner, repo, kanbanBoardName, b) {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
 func checkTargetBoard(owner, repo, kanbanBoardName string, b *trello.Board) bool {
-	if b.Name == kanbanBoardName && b.Desc == boardDesc(owner, repo) {
+	if !b.Closed && b.Name == kanbanBoardName && b.Desc == boardDesc(owner, repo) {
```
daniel-hutao, line 97: Just:
```go
return !b.Closed && b.Name == kanbanBoardName && b.Desc == boardDesc(owner, repo)
```

daniel-hutao: Maybe this function could be deleted for it only has one-line logic.

daniel-hutao (changes requested):
See the comment above.

docs/plugins/trello-github-integ_plugin.md
IronCore864: Good point. Because trello-github-integ plugin can only depend on Trello plugin.

internal/pkg/statemanager/manager.go
lfbdev: value, ok can only be used in if{} structure,  so if like this format,  will add "var value interfase{}" define

internal/pkg/plugin/trellogithub/helper.go
lfbdev: (tg *TrelloGithub) buildReadState() (map[string]interface{}, error) 

it need to fetch workflow path from remote:

path, err := tg.client.GetWorkflowPath()

internal/pkg/pluginengine/change.go
lfbdev: will print the options without true values

image

internal/pkg/pluginengine/getOutputs_test.go
lfbdev: will do it in an other PR for refactor

internal/pkg/pluginengine/fillRefValueWithOutputs_test.go
lfbdev: will do it in an other PR for refactor

internal/pkg/pluginengine/change_helper.go
lfbdev: logic is a bit complicated, it has been explained

internal/pkg/pluginengine/change.go
lfbdev: logic is a bit complicated, it has been explained

internal/pkg/statemanager/manager.go
daniel-hutao:
```go
value, ok := state.Resource["outputs"]
if !ok {
return nil, fmt.Errorf("cannot find outputs from state: %s", state.Name)
}
return value, nil
```

It seems easier to put the normal return value at the end

daniel-hutao:
Please don't forget to clear the commits record, then `push -f`, because we are no longer squash online.

IronCore864 (approved):

daniel-hutao (approved):

Issue #320, opened 2022-03-14
proposal: let dtm recognize that plugins version `exactly` match their own version


daniel-hutao:
## Description

Due to problems with the `golang plugins` mechanism, `dtm` and `plugins` must be compiled at the same time to ensure correct operation. For example, if the `dtm` is updated, the old `plugins` may not match. If the user builds the `dtm` by himself, after executing `dtm init`, the plugins with mismatched versions will be downloaded, and then strange bugs will be encountered.

eg.

```sh
2022-03-14 22:27:28 ℹ [INFO]  go-webapp-repo-Create -> plugin.Open(".devstream/github-repo-scaffolding-golang-darwin-amd64_0.2.0"): plugin was built with a different version of package internal/abi
```

Thanks @summingyu for exposing this bug. Although it took me more than 1h to locate.

PR #337, opened 2022-03-21
Feat: version improvement

lfbdev:
## Description

1. if local plugin exists, validate its md5 with .md5, if mismatch, cannot use this plugin

2. if local plugin does not exist, download plugin from remote, and validate its md5 with  .md5, if mismatch, cannot use this plugin

3. when do init,  check dtm md5 with remote md5 first, if does not match, cannot continue

### Related Issues

#320 

### Screenshots
md5 match:
image

md5 mismatch:
image


Makefile
```diff
@@ -4,31 +4,87 @@ VERSION=0.3.0
...
+	cat .devstream/jira-github-integ-${GOOS}-${GOARCH}_${VERSION}.md5 >> .devstream/sum.md5
+
 	go build -buildmode=plugin -trimpath -gcflags="all=-N -l" -o .devstream/openldap-${GOOS}-${GOARCH}_${VERSION}.so ./cmd/openldap/
-	go build -trimpath -gcflags="all=-N -l" -ldflags "-X github.com/merico-dev/stream/cmd/devstream/version.Version=${VERSION}" -o dtm-${GOOS}-${GOARCH} ./cmd/devstream/
+	${MD5} .devstream/openldap-${GOOS}-${GOARCH}_${VERSION}.so > .devstream/openldap-${GOOS}-${GOARCH}_${VERSION}.md5
+	cat .devstream/openldap-${GOOS}-${GOARCH}_${VERSION}.md5 >> .devstream/sum.md5
+	data=shell awk BEGIN{RS=EOF}'{gsub(/\n/,":");print}' .devstream/sum.md5
+	${SED} 's/  /_/g' .devstream/sum.md5
+
+build: fmt vet build-plugin ## Build dtm & plugins locally.
```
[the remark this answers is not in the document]
lfbdev: Updated

cmd/devstream/version.go
```diff
@@ -16,5 +16,5 @@ var versionCMD = &cobra.Command{
 }
 
 func versionCMDFunc(cmd *cobra.Command, args []string) {
-	fmt.Println(version.Version)
+	fmt.Println("Version:", version.Version, "\nSupported plugins MD5:", version.MD5String)
```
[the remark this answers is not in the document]
lfbdev: Updated

cmd/devstream/version/helper.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"crypto/md5"
+	"fmt"
+	"io"
+	"os"
+)
+
+// CalcFileMD5 calculate file md5
+func CalcFileMD5(filename string) (string, error) {
```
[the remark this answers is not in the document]
lfbdev: done

README.md
```diff
@@ -107,6 +107,10 @@ For other plugins, checkout the [docs/plugins](./docs/plugins/) folder for detai
 
 ## Run
 
+_Please make sure the following tips before running_
```
IronCore864, line 110: I suggest we remove this part for two reasons:
- `dtm init` will tell the users exactly this
- we might need to add more info on this topic in other docs

cmd/devstream/init.go
```diff
@@ -30,3 +43,35 @@ func initCMDFunc(cmd *cobra.Command, args []string) {
 
 	log.Success("Initialize finished.")
 }
+
+// compareDtmMD5 compare dtm itself and remote md5, if does not match, cannot  download plugins from remote
```
IronCore864, line 47: Don't put a lot of code in the application directory. If you think the code can be imported and used in other projects, then it should live in the /pkg directory. If the code is not reusable or if you don't want others to reuse it, put that code in the /internal directory. Read our [Project Layout](https://github.com/merico-dev/stream/blob/main/docs/project_layout.md#cmd) doc again.

cmd/devstream/version.go
```diff
@@ -16,5 +16,5 @@ var versionCMD = &cobra.Command{
 }
 
 func versionCMDFunc(cmd *cobra.Command, args []string) {
-	fmt.Println(version.Version)
+	fmt.Println("Version:", version.Version)
```
IronCore864, line 19: Do not add "Version:"; the reason is already explained in the previous comment.

docs/best_practices/gitops.md
```diff
@@ -77,6 +77,7 @@ dtm init -f config.yaml
 ```
 
 This step downloads the required plugins according to the config file.
+If your dtm does not match the remote dtm, you will not be able to download plugins.
```
IronCore864, line 80: I do not think this is the proper doc to put this information.

docs/quickstart_en.md
```diff
@@ -58,7 +58,9 @@ and you should see similar output to:
 2022-03-04 12:08:20 ✔ [SUCCESS]  Initialize finished.
 ```
 
-This step downloads the required plugins according to the config file.
+- This step downloads the required plugins according to the config file.
```
IronCore864, line 61: Copy and paste my wording for this part:

---

This step verifies the MD5 sum of your `dtm` binary, downloads the required plugins according to the config file, and verifies the plugins' MD5 sums as well.

_Note: if your `dtm` binary's MD5 sum doesn't match the MD5 sum from our release page, `dtm init` will stop. If your local `dtm` MD5 differs, it indicates that you built the binary yourself (for developing purposes, for example). Due to the nature of the Go plugin, `dtm` must be built together with the corresponding plugins. So, if you are building `dtm`, you should also build the plugins as well, in which case, you do not need to run `dtm init` to download the plugins._

docs/quickstart_zh.md
```diff
@@ -60,7 +60,9 @@ dtm init -f config.yaml
 2022-03-04 12:08:20 ✔ [SUCCESS]  Initialize finished.
 ```
 
-这一步会根据配置文件来下载需要的插件。
+- 这一步会根据配置文件来下载需要的插件。
+- 如果dtm和远端的dtm不一致，将不会下载插件。
```
IronCore864, line 64: Here it's not very accurate. Translate the English version from the previous comment.

internal/pkg/pluginmanager/manager.go
```diff
@@ -6,10 +6,12 @@ import (
...
 	"path/filepath"
 
-	"github.com/spf13/viper"
-
+	"github.com/merico-dev/stream/cmd/devstream/version"
 	"github.com/merico-dev/stream/internal/pkg/configloader"
 	"github.com/merico-dev/stream/pkg/util/log"
+	md5helper "github.com/merico-dev/stream/pkg/util/md5"
+
+	"github.com/spf13/viper"
```
IronCore864, line 14: The order is not correct. Third-party libs before our own packages.

internal/pkg/pluginmanager/manager.go
```diff
@@ -25,37 +27,49 @@ func DownloadPlugins(conf *configloader.Config) error {
...
+			// check if the downloaded plugin md5 matches with .md5
+			if err := checkPluginMismatch(pluginDir, pluginFileName, pluginMD5FileName, tool.Name); err != nil {
 				return err
 			}
 			continue
 		}
-		// check md5
-		dup, err := checkFileMD5(filepath.Join(pluginDir, pluginFileName), dc, pluginFileName, tool.Plugin.Version)
+
+		// if .so exists
```
IronCore864, line 48: What if *.so exists but *.md5 doesn't?

internal/pkg/pluginmanager/manager.go
```diff
@@ -25,37 +27,49 @@ func DownloadPlugins(conf *configloader.Config) error {
...
+		// if .so exists, and matches with .md5, continue
+		if isMD5Match {
 			log.Infof("Plugin: %s already exists, no need to download.", pluginFileName)
 			continue
 		}
-
-		log.Infof("Plugin: %s changed and will be overrided.", pluginFileName)
-		if err = os.Remove(filepath.Join(pluginDir, pluginFileName)); err != nil {
+		// if .so exists, but mismatches with .md5, re-download
+		log.Infof("Plugin: %s doesn't match with dtm core and will be downloaded.", pluginFileName)
```
IronCore864, line 59: The log message is not accurate. The plugin doesn't match its own MD5 sum, not with dtm core. Here it has nothing to do with the core, right?

internal/pkg/pluginmanager/manager.go
```diff
@@ -25,37 +27,49 @@ func DownloadPlugins(conf *configloader.Config) error {
 
 	for _, tool := range conf.Tools {
 		pluginFileName := configloader.GetPluginFileName(&tool)
+		pluginMD5FileName := configloader.GetPluginMD5FileName(&tool)
+		// plugin does not exist
 		if _, err := os.Stat(filepath.Join(pluginDir, pluginFileName)); errors.Is(err, os.ErrNotExist) {
-			// plugin does not exist
-			err := dc.download(pluginDir, pluginFileName, tool.Plugin.Version)
-			if err != nil {
+			// download .so file
```
IronCore864, line 33: This part is duplicated code as the redownloadPlugins function.

internal/pkg/pluginmanager/manager.go
```diff
@@ -66,29 +80,63 @@ func CheckLocalPlugins(conf *configloader.Config) error {
...
+			return err
+		}
+	}
 	return nil
 }
 
-func checkFileMD5(file string, dc *PbDownloadClient, pluginFileName string, version string) (bool, error) {
-	localmd5, err := LocalContentMD5(file)
+// checkPluginMismatch check if the plugins match with .md5
+func checkPluginMismatch(pluginDir, soFileName, md5FileName, tooName string) error {
```
IronCore864, line 105: 1. Name: maybe `checkPluginMD5Mismatch`? `checkPluginMismatch` doesn't sound right, mismatch with what?
2. Maybe the func should return a boolean and error, instead of only error. In that case, we can write:
```go
match, err := pluginAndMD5Matches()
if err {xxx}
if !match {yyy}
```

internal/pkg/pluginmanager/manager.go
```diff
@@ -66,29 +80,63 @@ func CheckLocalPlugins(conf *configloader.Config) error {
...
+	}
+	// download .md5 file
+	if err := dc.download(pluginDir, pluginMD5FileName, version); err != nil {
+		return err
+	}
+	return nil
+}
+
+// DownloadDtmMD5 download remote dtm .md5 for compare with local .md5
+func DownloadDtmMD5(remoteMD5Dir, dtmFileName string) error {
```
IronCore864, line 136: DtmMD5 looks funny. Either DownloadDTMsMD5/DownloadMD5ofDTM, or DownloadDtmMd5 (for acronyms, either all upper, or all only initial upper).

internal/pkg/pluginmanager/manager.go
```diff
@@ -66,29 +80,63 @@ func CheckLocalPlugins(conf *configloader.Config) error {
...
+	dc := NewPbDownloadClient()
```
IronCore864, line 137: What does PB mean? I forgot. Maybe we need a better naming here.

internal/pkg/pluginmanager/pluginmd5_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/merico-dev/stream/internal/pkg/configloader"
+	md5helper "github.com/merico-dev/stream/pkg/util/md5"
+)
+
+// TestCheckLocalPlugins test plguin .so matches dtm core md5
+func TestCheckLocalPlugins(t *testing.T) {
```
IronCore864, line 16: I think this is more of a functional test, not unit test. For functional tests, we are already using the ginkgo framework. Please use ginkgo. If it's only unit tests, Go's "testing" is fine.

IronCore864: It's also good to add unit test for functions like checkPluginMismatch.

pkg/util/md5/md5.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"io/ioutil"
+
+	"k8s.io/utils/strings"
+)
+
+const md5_length = 32
+
+// ValidateFileMatchMD5 check current PlugIn MD5 matches with .md5 file
+func ValidateFileMatchMD5(fileName, md5FileName string) (bool, error) {
```
IronCore864, line 12: Name: FileMatchesMD5 sounds better, because then we can use it like:
```go
if match, err := FileMatchesMD5()
...
```

If the function name is like "validate file md5" or something like that, I would expect that the first param is the filename or content, and the second is the md5 sum, not a file of md5. So, maybe (filename, md5sum string) is better?

pkg/util/md5/md5.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"io/ioutil"
+
+	"k8s.io/utils/strings"
+)
+
+const md5_length = 32
+
+// ValidateFileMatchMD5 check current PlugIn MD5 matches with .md5 file
```
IronCore864, line 11: ValidateFileMatchMD5 check -> ValidateFileMatchMD5 checks.

Please also check other comments.

internal/pkg/pluginmanager/manager.go
lfbdev: PbDownloadClient， it means  Progress bar downloader

cmd/devstream/init.go
lfbdev: yes，I will move them into dir /pkg/md5

cmd/devstream/version.go
lfbdev: This version is for "dtm version",  and also this Version is used in other files instead of "plugin.version"

lfbdev: yes,  got it

internal/pkg/pluginmanager/manager.go
lfbdev: updated,
if no *.so, download it;
if no *.md5, download it;
then compare them;
if match, contiue;
if not match, re-download(remove and download both .so and .md5)
comapre them
if not match, raise error

internal/pkg/pluginmanager/manager.go
lfbdev: 1. I changed  the function name
2. It wraps function ValidateFileMatchMD5, not return bool, in order to throw the result of the mismatch as an error, thereby ending the process

internal/pkg/pluginmanager/manager.go
lfbdev: updated

pkg/util/md5/md5.go
lfbdev: updated

internal/pkg/pluginmanager/manager.go
lfbdev: updated.
Here are the downloads separately, .so and .md5 whichever is missing, download whichever one is missing.
redownloadPlugins  function removes both .so and .md5 and download

internal/pkg/pluginmanager/pluginmd5_test.go
lfbdev: I think CheckLocalPlugins(config) is also a function, just like others

docs/quickstart_zh.md
lfbdev: Updated

internal/pkg/pluginmanager/pluginmd5_test.go
IronCore864: I think the test for "FileMatchesMD5" is a "unit" test (technically speaking, it isn't, since it has dependencies on the os.Create, but at least the function is relatively isolated without other dependencies).

The test for CheckLocalPlugins, on the other hand, depends on creating a config, calling configloader's functions to prepare for the test, etc., which are some external dependencies. Yes, technically speaking, the test for "CheckLocalPlugins" isn't "functional" test either, but given the complexity of building the dependencies and stuff, I think we would benefit from using ginkgo.

README.md
```diff
@@ -156,6 +156,7 @@ To destroy everything, run:
 
 _`dtm` will read the state, then determine which tools are installed, and then remove those tools. It's same as `dtm apply -f empty.yaml` (empty.yaml is an empty config file)._
 
+
```
IronCore864, line 159: It seems here we have added an extra new line for no reason.

docs/best_practices/gitops.md
```diff
@@ -76,8 +76,6 @@ Run:
 dtm init -f config.yaml
 ```
 
-This step downloads the required plugins according to the config file.
-
```
IronCore864, line 80: We don't need to delete these two lines.

internal/pkg/pluginmanager/manager.go
```diff
@@ -25,37 +30,48 @@ func DownloadPlugins(conf *configloader.Config) error {
...
+			if err := dc.download(pluginDir, pluginFileName, tool.Plugin.Version); err != nil {
 				return err
 			}
-			continue
+			log.Successf("[%s] download succeeded.", pluginMD5FileName)
 		}
-		// check md5
-		dup, err := checkFileMD5(filepath.Join(pluginDir, pluginFileName), dc, pluginFileName, tool.Plugin.Version)
+		// .md5 does not exist
+		if _, err := os.Stat(filepath.Join(pluginDir, pluginFileName)); errors.Is(err, os.ErrNotExist) {
```
IronCore864, line 43: pluginFileName here should be pluginMD5FileName?

internal/pkg/pluginmanager/pbdownloader.go
```diff
@@ -36,6 +36,8 @@ func (pd *PbDownloadClient) download(pluginsDir, pluginFilename, version string)
 	}
 
 	downloadURL := fmt.Sprintf("%s/v%s/%s", defaultReleaseUrl, version, pluginFilename)
+	log.Debugf("downloading url is: %s.", downloadURL)
```
IronCore864, line 39: Uppercase D

cmd/devstream/init.go
```diff
@@ -22,6 +22,11 @@ func initCMDFunc(cmd *cobra.Command, args []string) {
 	}
 
 	log.Info("Initialize started.")
+	if err := pluginmanager.CompareDtmMD5(); err != nil {
+		log.Error(err)
```
IronCore864, line 26: "err" is generated by errorf(), which results in a sentence fragment, not a full sentence. However, log should be a full sentence. So, here it will not print pretty logs. We should do something like "Got error during xxx, err: %s" (for example, don't copy paste).

internal/pkg/md5/helper.go
```diff
@@ -0,0 +1,25 @@
+package md5
```
daniel-hutao, line 1: package `md5` should be under `pkg/util/`

internal/pkg/pluginmanager/manager.go
```diff
@@ -4,14 +4,19 @@ import (
...
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 
 	"github.com/spf13/viper"
 
+	"github.com/merico-dev/stream/cmd/devstream/version"
 	"github.com/merico-dev/stream/internal/pkg/configloader"
+	md5helper "github.com/merico-dev/stream/internal/pkg/md5"
```
daniel-hutao, line 14: why rename the md5 to md5helper?

internal/pkg/pluginmanager/manager.go
```diff
@@ -4,14 +4,19 @@ import (
...
 	"github.com/merico-dev/stream/pkg/util/log"
 )
 
+const DTMRemoteMD5 = ".remote"
```
daniel-hutao, line 18: The const name `DTMRemoteMD5` seems hard to understand, why is the remote md5 a string? Or conversely, what exactly does this string .remote mean?

internal/pkg/pluginmanager/manager.go
```diff
@@ -66,29 +82,98 @@ func CheckLocalPlugins(conf *configloader.Config) error {
...
 			if err != nil {
 				return err
 			}
 			return fmt.Errorf("plugin %s doesn't exist", tool.Name)
 		}
+		if _, err := os.Stat(filepath.Join(pluginDir, pluginMD5FileName)); errors.Is(err, os.ErrNotExist) {
+			if err != nil {
+				return err
+			}
+			return fmt.Errorf(".md5 file of plugin %s doesn't exist", tool.Name)
```
daniel-hutao, line 96: ? why return the err when `err != nil` and return another err when `err==nil`

internal/pkg/pluginmanager/manager.go
```diff
@@ -66,29 +82,98 @@ func CheckLocalPlugins(conf *configloader.Config) error {
...
+}
+
+// pluginAndMD5Matches checks if the plugins match with .md5
+func pluginAndMD5Matches(pluginDir, soFileName, md5FileName, tooName string) error {
+	isMD5Match, err := md5helper.FileMatchesMD5(filepath.Join(pluginDir, soFileName), filepath.Join(pluginDir, md5FileName))
+	if err != nil {
+		return err
+	}
+	if !isMD5Match {
+		return fmt.Errorf("plugin %s doesn't match with .md5", tooName)
```
daniel-hutao, line 113: if err == nil, then return another err again?

daniel-hutao (changes requested):

internal/pkg/pluginmanager/manager.go
lfbdev: It is a dir name,  I renamed it to "DTMRemoteMD5Dir"

internal/pkg/pluginmanager/manager.go
lfbdev: err == nil && isMD5Match， return nil
err == nil && !isMD5Match， return fmt.Errorf("plugin %s doesn't match with .md5", tooName)

internal/pkg/pluginmanager/manager.go
lfbdev: updated

internal/pkg/pluginmanager/manager.go
lfbdev: updated

internal/pkg/pluginmanager/manager.go
lfbdev: This is left over from the original code, I changed it

cmd/devstream/init.go
lfbdev: Updated

IronCore864 (approved):

daniel-hutao (approved):

PR #346, opened 2022-03-24
refactor(trello): consistent entry logic with other plugins

daniel-hutao:

## Description

Consistent entry logic with other plugins


internal/pkg/plugin/trello/create.go
```diff
@@ -1,27 +1,32 @@
...
 )
 
 // Create creates Tello board and lists(todo/doing/done).
 func Create(options map[string]interface{}) (map[string]interface{}, error) {
-	var opts *Options
-	var err error
-
-	if opts, err = convertMap2Options(options); err != nil {
+	var opts Options
+	if err := mapstructure.Decode(options, &opts); err != nil {
```
IronCore864, line 14: Maybe we can add it into some document to let future developers know which function to use to decode options.

IronCore864 (approved):

internal/pkg/plugin/trello/read.go
```diff
@@ -1,17 +1,25 @@
...
 		return nil, err
 	}
 
-	if err := validateOptions(opts); err != nil {
-		return nil, err
+	if errs := validate(&opts); len(errs) != 0 {
+		for _, e := range errs {
+			log.Errorf("Options error: %s.", e)
+		}
+		return nil, fmt.Errorf("opts are illegal")
```
lfbdev, line 21: opts and Options are not consistent

lfbdev (approved):

internal/pkg/plugin/trello/create.go
daniel-hutao: Get it

internal/pkg/plugin/trello/read.go
daniel-hutao: I write this because other plugins have been written this way before, and consistency is the main consideration here. Of course, it may be more appropriate to change to `options` here, but all plugins need to be changed together. Maybe I'll make adjustments later when I add this logic to "plugin template".

Issue #342, opened 2022-03-23
Proposal: Refactor All Plugins to Be Consistent with The `Plugin Template`


daniel-hutao:
## Description

Now we have tools to automatically generate the scaffolding of a plugin. But we haven't before.

Yes, so there must be some inconsistencies in so many plugins that have been implemented before. 

Let me think, how can we fix it?

## How to do it

First, you need to know how we generate a plugin scaffolding now.
You could see the #325 for more info with the command `dtm develop create-plugin --name=foo`.

Of course, you'll need to try running this command locally to see what it scaffold for you. You can see the local changes with the `git` tool. Going a step further, maybe you can take a look at the corresponding source code of `develop` sub-command, maybe all the details you need to know can be obtained this way.

Then, you know what materials a new plugin contains. Yes, the next step is to see how the old plugins differs from the new one and try to make that drift go away.

Of course, the `dtm develop create-plugin --name=foo` command hasn't been practiced, so it's not necessarily right. So, stay skeptical and give feedback if you feel there is anything unreasonable about this command.

## Last but not least

Have fun with refactoring. Don't be shy and tell me if you need help, I'm always there.

imxw:
> 

I'm happy to do that. Just waiting for the distribution of @daniel-hutao

daniel-hutao:
ok，ok，ok，3 volunteers now，，，

daniel-hutao:
@imxw 
- [ ] argocd
- [ ] kube-prometheus
- [ ] jenkins 
- [ ] openldap 
- [ ] githubactions-golang

@warren830 
- [ ] argocdapp
- [ ] devlake
- [ ] trello
- [ ] gitlabci  
- [ ] githubactions-nodejs


@summingyu 
- [ ] github-repo-scaffolding-golang
- [ ] jira-github
- [ ] trello-github
- [ ] githubactions-python

daniel-hutao:
#345

PR #359, opened 2022-03-27
refactor(plugin): adapt for the plugin generation tool

imxw:
## Description

Adjust the following plugins format to fit the Tool Builder format

- [x]  argocd
- [x] kube-prometheus
- [x] jenkins
- [x] openldap
- [x] githubactions-golang

### Related Issues

#342


internal/pkg/plugin/githubactions/golang/validate.go
```diff
@@ -0,0 +1,73 @@
...
+	if errs := opts.Docker.Validate(); len(errs) != 0 {
+		for _, e := range errs {
+			retErrors = append(retErrors, fmt.Errorf("docker is invalid: %s", e))
+		}
+	}
+
+	return retErrors
+}
+
+func parseAndValidateOptions(options map[string]interface{}) (*Options, error) {
```
daniel-hutao, line 58: This function should not exist because it makes the entry of the corresponding plugins inconsistent with the template (including most other plugins).

- template/create.go
 
```go
func Create(options map[string]interface{}) (map[string]interface{}, error) {
	var opts Options
	if err := mapstructure.Decode(options, &opts); err != nil {
		return nil, err
	}

	if errs := validate(&opts); len(errs) != 0 {
		for _, e := range errs {
			log.Errorf("Options error: %s.", e)
		}
		return nil, fmt.Errorf("opts are illegal")
	}

    // TODO(dtm): Add your logic here.

    return nil, nil
}
```

/cc @warren830 @summingyu Please also pay attention to this tip.

daniel-hutao (changes requested):
See the comments.

internal/pkg/plugin/githubactions/golang/create.go
```diff
@@ -10,11 +13,20 @@ import (
 
 // Create sets up GitHub Actions workflow(s).
 func Create(options map[string]interface{}) (map[string]interface{}, error) {
-	opts, err := parseAndValidateOptions(options)
-	if err != nil {
+
```
daniel-hutao, line 16: unnecessary blank line

daniel-hutao (commented):
see the comment

daniel-hutao (approved):
/lgtm

IronCore864 (approved):

PR #360, opened 2022-03-27
Automated Release script and workflow

lfbdev:
# Summary

Background:
trigger can be manual, build/upload for 3 platforms should be automatic

add auto release script for darwin arm64
add githubactions for linux and darwin amd64

## Description

using github.com/github-release/github-release for uploading assets

### Related Issues

#364 

### Current Behavior

### New Behavior
image

image


.github/workflows/automated-release.yml
```diff
@@ -0,0 +1,116 @@
...
+      TAG: ${{ github.ref_name }}
+      USER: merico-dev
+      REPO: stream
+      PLUGINDIR: .devstream
+      GOOS: linux
+      GOARCH: amd64
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "🐧 This job is now running on a ${{ runner.os }}-${{ runner.arch }} server hosted by GitHub!"
+      - name: Checkout
```
IronCore864, line 18: I think we should explicitly check out the tag. I don't know if it's possible, but what if I publish a release then immediately a new commit is generated by others? In that case, will it check out the latest commit instead of this tag?

.github/workflows/automated-release.yml
```diff
@@ -0,0 +1,116 @@
...
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+      - name: Build
+        run: make build
+      - name: Install Github-release
+        run: go install github.com/github-release/github-release@latest
+      - name: upload
+        run: |
+          #upload dtm
```
IronCore864, line 39: I suggest we create a script for this part, so that it's easier to manage.

build/package/auto-release-darwin-arm64.sh
```diff
@@ -0,0 +1,46 @@
...
+   read_dir $plugin_dir"/"$file
+ else
+   echo 'Uploading '$file' ...'
+   github-release upload --security-token $github_token --user $user --repo $repo --tag $tag --file  $plugin_dir"/"$file --name $file
+   echo $file' uploaded.'
+ fi
+done
+}
+
+#upload dtm
```
IronCore864, line 35: Please pay attention to the comment format. Keep a space after the hashtag.

daniel-hutao (commented):
I think it's better to save the `build/package/auto-release-darwin-arm64.sh` to `hack/release/` and then delete the `package` dir.

.github/workflows/automated-release.yml
lfbdev: ok, will refactor it

lfbdev:
build/package/ contains other files, such as build_linux_amd64.sh and Dockerfile

daniel-hutao:
> 

Yes, so what I mean is delete all files under the `build/package`. They are no use after release with gh-actions.

.github/workflows/automated-release.yml
lfbdev: yes, i added a option for checkout the specified tag.

IronCore864 (approved):

daniel-hutao (approved):

Issue #326, opened 2022-03-16
proposal: trigger GitHub actions after some specific comments


daniel-hutao:
## Description

Currently, we trigger actions through events such as `pr created/updated` which is obviously a waste of ci bandwidth because pr may be updated frequently. 

I think it is more controllable, elegant, and efficient to trigger the ci process through comments.

/co-proposed @IronCore864 

## Describe the Proposal

When a pr is created, the ut actions flow can be triggered by the comment as below:

- `/ok-to-test` runs UT in Linux(default os env)
- `/test-in-mac` runs UT in Mac

e2e/build/lint/.. Other workflows are not without design, but it does not need to be done here. When the ut process goes through this pattern, I will write a detailed document to record all the supported comments and write the corresponding xxx-actions.yml.

@IronCore864 @lfbdev Please help to take a look at this proposal and give me some advice.

---

We are gradually implementing some kind of ChatOps. `Comment` is one, `IM(Feishu)` is one, isn't it interesting? I can't help but(忍不住) tell you, let's play together @imxw @warren830 @summingyu

PR #362, opened 2022-03-28
feat: trigger GitHub actions after some specific comments

daniel-hutao:

# Summary

feat: trigger GitHub actions after some specific comments
/area ci

## Description

If a `pr`(a kind of issue) is created or updated, `build and test` will be triggered if the comment contains "/ok-to-test"

### Related Issues

#326 

### Screenshots

1. pr created without comment `/ok-to-test`
image

### TO-DO

- [ ] Adding the triggering personnel limit - specify that it can only be triggered by members of the `devstream-io/developer` team.
- [ ] It may be necessary to distinguish between mac and linux, but it seems that there is no need to do a fine split at present, which may be counterproductive.


daniel-hutao:
image

I've deleted the `pr created` and added the `ut on linux`, but the `GitHub Actions` still run `pr created` stage. Maybe it's a bug with `GitHub Actions`.

I tested the `ok-to-test` at my own repo and it is work fine:
image

(Please ignore the error below, It's triggered by the Makefile missing)
image

lfbdev (approved):

IronCore864 (approved):

PR #365, opened 2022-03-28
chore: remove `Foo string` field

imxw:
# Summary

`Foo string` field is not necessary in the `options.go` template


daniel-hutao (commented):
It's a sample field config for users. So I think it's ok to keep it.

imxw:
> It's a sample field config for users. So I think it's ok to keep it.

Got it!

PR #376, opened 2022-03-30
fix: fix the return error

lfbdev:
# Summary

should be if... else return err

### DIFF
--- a/internal/pkg/pluginmanager/manager.go
+++ b/internal/pkg/pluginmanager/manager.go
@@ -33,27 +33,25 @@ func DownloadPlugins(conf *configloader.Config) error {
 		pluginMD5FileName := configloader.GetPluginMD5FileName(&tool)
 		// plugin does not exist
 		if _, err := os.Stat(filepath.Join(pluginDir, pluginFileName)); err != nil {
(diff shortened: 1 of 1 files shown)


internal/pkg/pluginmanager/manager.go
```diff
@@ -39,8 +39,9 @@ func DownloadPlugins(conf *configloader.Config) error {
 					return err
 				}
 				log.Successf("[%s] download succeeded.", pluginFileName)
+			} else {
```
daniel-hutao, line 42:
```go
		if _, err := os.Stat(filepath.Join(pluginDir, pluginFileName)); err != nil {
			if !errors.Is(err, os.ErrNotExist) {
				return err
			}
			// download .so file
			if err := dc.download(pluginDir, pluginFileName, version.Version); err != nil {
				return err
			}
			log.Successf("[%s] download succeeded.", pluginFileName)
		}
```

too many `if`

daniel-hutao (changes requested):

daniel-hutao (approved):

IronCore864 (approved):

PR #385, opened 2022-04-04
doc: installation on mac

algobot76:
# Summary

## Key Points

- [ ] This is a breaking change.
- [x] Documentation (new or existing) is updated.

## Description

- Doc about installation instructions.
- A few fixes, including deleting extra whitespace and adding missing empty lines.

### Related Issues

Close #372 

### Current Behavior

n/a

### New Behavior

n/a

### Screenshots

n/a

### Other Information

n/a

PR #386, opened 2022-04-04
fix: golang lint ci failed

daniel-hutao:

# Summary

fix: golang lint ci failed

## Description

Just now I found out that one of the contributor's pr(#385 ) failed in ci workflow:
https://github.com/devstream-io/devstream/runs/5814780108?check_suite_focus=true

This problem is not introduced by this pr, but: https://github.com/golangci/golangci-lint-action/issues/434

So the solution here should be:

1. set up `skip-go-installation: true` in v2
2. upgrade `golangci-lint-action` to v3 - (I choose this)

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -13,9 +13,13 @@ jobs:
     name: Golang lint
     runs-on: ubuntu-latest
     steps:
+      - name: Install Go 1.17
+        uses: actions/setup-go@v2
+        with:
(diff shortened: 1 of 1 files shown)


daniel-hutao:
/cc @devstream-io/pmc 
In order not to block the contributors from run the ci workflows, and not to disturb everyone's vacation, I merged this pr first while ensuring that the test passed. Of course, if a new problem is caused, I will fix it ASAP.

Issue #383, opened 2022-04-02
Proposal: I would like to have support for generating the default config configuration using the command


summingyu:
## Description

I would like to have support for generating the default config configuration using the command
such as `dtm config githubaction ` to generate the default configuration of githubaction

## Describe the Proposal

The current way to get the plugin configuration file is to visit the docs directory of the github repository.
However, if a customer installs with brew or has only one binary, they may not know how the plugin configuration file is configured.
So there is a need to have a convenient way to show the default configuration of plugin.

Or go a step further and generate the toolschain configuration file with a command.
For example `dtm config toolschain p1-p2-p3`

daniel-hutao:
Great idea! In fact, as early as a few months ago, I thought about generating a specific devops toolchain through the command line interactively, such as the `GitOps chain`. 
I have also discussed this idea with @IronCore864  before, but since `DevStream` is not mature enough, the idea had been waiting for the right time.

---

We can implement this idea step by step. In order to avoid getting into complications at the beginning, we can first implement such a feature: 
**output the default configuration of any plugin through the command line**, 
the output is exactly the same as the default configuration written in that plugin's documentation. In this way the user does not need to read the documentation to learn the configuration of a plugin. When this feature is perfected, we further think about the more interesting feature later.

---

Then I have a simple design for your reference:
- command: `dtm config default --plugin=A_PLUGIN_NAME`.
- All plugins need to support this command.
- Maybe add an `output.go` file in the `internal/pkg/plugin/PLUGIN_NAME/` directory to put this the "output content"
- The "output content" may be like the code below:

```go
var DefaultConfigContent = `tools:
- name: NAME
  # name of the plugin
  plugin: NAME
  options:
    foo: bar
`

```
- The generation with `output.go` template should be adding to `dtm develop xxx` command.
- Of course, related documents also need to be updated, but we can wait until this ticket is merged and then update docs separately. Because the document may involve multiple places such as repo/blogs, it will also involve some additional thinking, such as whether to update the old blog, or add a new blog, and then add the link to the old blog. In short, documentation is a separate piece of work but important too.

/cc @devstream-io/devstream

daniel-hutao:
@summingyu Please wait for @IronCore864 's reply, after determine the implementation method, then start development.

IronCore864:
@summingyu great idea!

[...]

PR #392, opened 2022-04-07
Feat: support "default config print" logic

daniel-hutao:

# Summary

Feat: support "default config print" logic

- support `default config` when scaffolding a new plugin
- support default config print logic in `argocd`
- simplify the implementation with `dtm list plugins`

### Related Issues

#383 

### Screenshots

1. `./dtm develop create-plugin --name=foo`

image

2. `./dtm show config --plugin=argocd`

image

3. `./dtm show config --plugin=abcd`

image

4. `./dtm list plugins`

image


internal/pkg/show/config/plugin/argocd.go
```diff
@@ -0,0 +1,34 @@
...
+      # the time to wait for any individual Kubernetes operation (like Jobs for hooks). This defaults to 5m0s
+      timeout: 5m
+      # whether to perform a CRD upgrade during installation
+      upgradeCRDs: true
+      # custom configuration (Optional). You can refer to [ArgoCD values.yaml](https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/values.yaml)
+      values_yaml: |
+        controller:
+          service: 
+            port: 8080
+`
```
IronCore864, line 34: We should remove the "\n" after "8080", otherwise the output has an extra new line:

Screen Shot 2022-04-07 at 11 24 09

internal/pkg/show/config/config.go
```diff
@@ -0,0 +1,22 @@
...
+	"argocd": plugin.ArgocdDefaultConfig,
+}
+
+func Show() error {
+	plugin := viper.GetString("plugin")
+	if config, ok := pluginDefaultConfigs[plugin]; ok {
+		fmt.Println(config)
+		return nil
+	}
+	return fmt.Errorf("plugin name error: < %s >", plugin)
```
IronCore864, line 21: If we run `dtm show config` without plugin name, it will print a log like "plugin name error: <  >". This is not user-friendly. We should see if the "plugin" is empty or not, then give useful error messages. For example:

"Empty plugin name. Maybe you forgot to add --plugin=PLUGIN_NAME?"

cmd/devstream/develop.go
```diff
@@ -32,7 +32,7 @@ func developCMDFunc(cmd *cobra.Command, args []string) {
 	}
 }
 
-func validateArgs(args []string) error {
+func validateDevelopArgs(args []string) error {
 	// "create-plugin"/ maybe it will be "delete-plugin"/"rename-plugin" in future.
 	if len(args) != 1 {
 		return fmt.Errorf("got illegal args count (expect 1, got %d). "+
```
IronCore864, line 38: Let's refactor this error message.

It should not be a full sentence. "illegal args count", "missing arg(s)" is better, remove the verb "got".

We should not add another full sentence "See `help` command for more info".

It will look like "got illegal args count (expect 1, got 0). See `help` command for more info", it's not a full sentence as a log line.

cmd/devstream/show.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/devstream-io/devstream/internal/pkg/show"
+	"github.com/devstream-io/devstream/pkg/util/log"
+)
+
+var plugin string
+
+var showCMD = &cobra.Command{
+	Use:   "show",
```
IronCore864, line 15: Use: "show [plugin]"

This will print:

```shell
Usage:
  dtm show [plugin] [flags]
```

In the future, we could have:

```shell
Usage:
  dtm show [plugin|outputs|xxx] [flags]
```

cmd/devstream/show.go
```diff
@@ -0,0 +1,50 @@
...
+	Short: "Show is used to print some useful information",
+	Long: `Show is used to print some useful information. 
+eg.
+- dtm show config --plugin=A-PLUGIN-NAME`,
```
IronCore864, line 19: The correct abbreviation of "for example" is "e.g.", and since this is the beginning of a sentence, it should be:

"E.g.:"

But making the first letter uppercase looks weird, so maybe put "Examples:" instead.

Also, "-" can't be rendered as a bullet point like in markdown. so maybe we should use two spaces instead of a bullet.

```shell
Examples:
  dtm show config --plugin=argocd
```

IronCore864 (changes requested):

internal/pkg/show/config/plugin/argocd.go
daniel-hutao: got it

internal/pkg/show/config/config.go
daniel-hutao: got it

cmd/devstream/develop.go
daniel-hutao: return fmt.Errorf("illegal args count (expect 1, got %d)", len(args))

cmd/devstream/show.go
daniel-hutao: 👍

cmd/devstream/show.go
daniel-hutao: updated at list/show/develop

daniel-hutao:
@IronCore864 Please take a review again. All fixed.

PR #398, opened 2022-04-08
ci: optimize ci pipeline

imxw:
# Summary

## Description

- Merge some ci pipelines into one because of the order of precedence. For example, if `golang-lint` is failed, there is no need to continue with the remaining steps
- Add go build cache for MacOS (`~/Library/Caches/go-build`)
- Add file filtering so that only changes to these files will trigger the pipeline
- Add conditional judgment for `e2e-test`, only triggered if the repository is `devstream-io/devstream`

### Related Issues

N/A

### Screenshots

image

### Other Information
- The first time will be a bit slow because there is no cache
- Builds on MacOS occasionally have issues that can slow down the overall progress. If this is often the case later, it can be moved to another pipeline


.github/workflows/main.yml
```diff
@@ -0,0 +1,78 @@
...
+            ${{ matrix.go-build }}
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ matrix.go }}-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-${{ matrix.go }}-
+            ${{ runner.os }}-go-
+      - name: Test
+        run: go test -v ./...
+      - name: Build
+        run: make build -j8
```
daniel-hutao, line 78: missing new-line

.github/workflows/e2e-test.yml
```diff
@@ -3,28 +3,44 @@ on:
   push:
     branches:
       - main
+    paths:
+      - '**.go'
+      - 'Makefile'
+      - 'go.md'
```
daniel-hutao, line 9: go.mod?

imxw: yes

.github/workflows/main.yml
```diff
@@ -0,0 +1,78 @@
+name: main builder
```
daniel-hutao, line 1: If this file's name changed to `main.yml`, you need to update the reference at README.md(line 8) too. Also notice the README_zh.md

.github/workflows/main.yml
```diff
@@ -0,0 +1,78 @@
...
+
+on:
+  pull_request:
+    branches: [ main ]
+    paths:
+      - '**.go'
+      - 'Makefile'
+      - 'go.md'
```
daniel-hutao: go.mod?

.github/workflows/e2e-test.yml
```diff
@@ -3,28 +3,44 @@ on:
...
-            ~/.cache/go-build
             ~/go/pkg/mod
-          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+            ~/.cache/go-build
+          key: ${{ runner.os }}-go-${{ matrix.go }}-${{ hashFiles('**/go.sum') }}
           restore-keys: |
+            ${{ runner.os }}-go-${{ matrix.go }}-
             ${{ runner.os }}-go-
       - name: Build
         run: make build -j8
```
daniel-hutao, line 45: If we remove the `push to main` workflow, the ut will never run at main branch. Should we add the ut step here?

imxw: The e2e test is triggered when merging into the main branch

imxw:
I found that `lint-commit-message` still has to be split out, because `main.yml` sets up a file filter so that only go-related changes will trigger the pipeline

.github/workflows/e2e-test.yml
imxw: > If we remove the `push to main` workflow, the ut will never run at main branch. Should we add the ut step here?

`pull request` will trigger ut, how about having `main.yml` support push events?

daniel-hutao: Of course, after adding the `push events trigger`, this logic is complete. It's just that I see the simplification you have done. I guess it is to reduce the workflow on the main branch updated, so it seems easier to add `ut` to e2e.yml; if we add `push` directly to main.yml, it seems that `main` and `e2e` cannot avoid repeated `make build`.

.github/workflows/lint-commit-message.yml
```diff
@@ -1,13 +1,13 @@
...
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           fetch-depth: 0
       - uses: wagoid/commitlint-github-action@v4
         with:
-          configFile: ${{ github.workspace }}/build/ci/commitlint.config.js
+          configFile: ${{ github.workspace }}/build/ci/commitlint.config.js
```
IronCore864, line 13: There is a missing "end of file" at the end of the line. Maybe use an IDE plugin to solve this issue once and for all.

imxw: Got it, I will fix this issue

IronCore864 (approved):

imxw:
@daniel-hutao

daniel-hutao (approved):
/LGTM

Issue #380, opened 2022-04-02
Proposal: Add fig autocomplete support


imxw:
## Description

I've been using the command line for a long time, and I'm a user of **fig autocomplete** which is very cool. It lets you add visual autocomplete to any CLI tool or script. If `dtm` supports **fig autocomplete**, I believe it will greatly  improve the user experience.

## Describe the Proposal

See https://github.com/withfig/autocomplete

## Has the Feature been Requested Before?

No

## Describe the Alternatives You Have Considered

n/a

## Additional Context

<img width="533" alt="image" src="https://user-images.githubusercontent.com/16207029/161362949-cfe1a061-8763-4034-8716-94434776b1f9.png">

<img width="555" alt="image" src="https://user-images.githubusercontent.com/16207029/161362917-603befdf-1503-4208-b68a-f534d433ee77.png">

![test](https://user-images.githubusercontent.com/16207029/161363144-1dfc2ec6-c660-4f75-9207-596d15939a24.gif)

daniel-hutao:
This idea sounds cool!
I agree to add this feature.
I don't know what everyone thinks @devstream-io/pmc ;
If all agree, I suggest this feature can be assgin @imxw to realize.

IronCore864:
`fig` looks nice. Please also consider alternatives like `bash-completion`, examples:
- https://kubernetes.io/docs/tasks/tools/included/optional-kubectl-configs-bash-linux/
- https://kubernetes.io/docs/tasks/tools/included/optional-kubectl-configs-bash-mac/

Make a decision using your best judgement :)

daniel-hutao:
https://github.com/withfig/autocomplete/pull/1140

daniel-hutao:
A document with "how to use fig along with dtm" is needed. 

@imxw @IronCore864

imxw:
What to pay attention to when writing documents now?

PR #413, opened 2022-04-14
docs: add `fig` doc

imxw:
# Summary

## Key Points

- [x] Documentation (new or existing) is updated.

## Description

Add `fig` use documentation

### Related Issues

#380


docs/fig.md
```diff
@@ -0,0 +1,44 @@
+# Fig Auto-complete
```
daniel-hutao, line 1: Maybe the filename is better to use "fig-autocomplete.md", for many users know the "autocomplete", but don't know the fig. (maybe, just my personal guess)

docs/index.md
```diff
@@ -18,6 +18,7 @@ project_layout
 creating_a_plugin
 best_practices
 creating_a_document
+fig
```
daniel-hutao, line 21: Same as above

daniel-hutao (changes requested):

docs/fig.md
IronCore864: I agree. Maybe simply "autocomplete" without even mentioning "fig" in the doc's filename/title is also fine. Then in the first paragraph, we can mention that it's implemented by using xxx tool.

imxw: I will rename it to autocomplete and also mention other complete tools. WDYT?

daniel-hutao: > I agree. Maybe simply "autocomplete" without even mentioning "fig" in the doc's filename/title is also fine. Then in the first paragraph, we can mention that it's implemented by using xxx tool.

autocomplete is better. because we could add "linux support with other tool" in the same doc.

daniel-hutao (approved):
Good Job!

IronCore864 (approved):

PR #426, opened 2022-04-21
fix: validate params

lfbdev:
## Description

auto-release-darwin-arm64.sh should check all variables. If empty, no error should occur.

### Related Issues

#427 

### Current Behavior

image

### DIFF
--- a/build/package/auto-release-darwin-arm64.sh
+++ b/build/package/auto-release-darwin-arm64.sh
@@ -1,4 +1,5 @@
 #! /bin/bash -e
+set -o nounset
 
 tag=v0.4.0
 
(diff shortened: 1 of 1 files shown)


build/package/auto-release-darwin-arm64.sh
```diff
@@ -10,6 +10,16 @@ plugin_dir=.devstream
 GOOS=$(go env GOOS)
 GOARCH=$(go env GOARCH)
 
+if [ ! $tag ] || [ ! $user ] || [ ! $repo ] || [ ! $github_token ] || [ ! $plugin_dir ]; then
+  echo "The following variables cannot be empty!"
+  echo "tag="$tag
+  echo "user="$user
+  echo "repo="$repo
+  echo "github_token="$github_token
```
IronCore864, line 18: if github_token isn't empty but maybe for example, tag is empty, github_token will be printed out on the screen, which isn't safe.

daniel-hutao: +1

daniel-hutao (changes requested):
set -o errexit
set -o nounset
set -o pipefail

build/package/auto-release-darwin-arm64.sh
lfbdev: 1.  add "set -o nounset" for undefined parameter
2. do not  print the true token

daniel-hutao (approved):

PR #442, opened 2022-05-04
chore: optimize Makefile

imxw:
# Summary

## Description

- No more hard-coding VERSION variable in the `Makefile`. Don't need to modify the value of VERSION when you release `dtm`, just execute this command `make build VERSION=xxx`
- Use a more native way to get plugins list. `PLUGINS=$(notdir $(wildcard $(ROOT_DIR)/cmd/plugin/*))`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,24 +1,35 @@
-VERSION=0.4.0
 
-MKFILE_PATH=$(abspath $(lastword $(MAKEFILE_LIST)))
-BUILD_PATH=$(patsubst %/,%,$(dir $(MKFILE_PATH)))/build/working_dir
+SELF_DIR=$(dir $(lastword $(MAKEFILE_LIST)))
 
 GOOS=$(shell go env GOOS)
 GOARCH=$(shell go env GOARCH)
-GO_BUILD=go build -buildmode=plugin -trimpath -gcflags="all=-N -l"
-PLUGINS_CMD_ROOT=./cmd/plugin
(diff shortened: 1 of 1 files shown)


IronCore864:
I have only one small concern: we run `make build` a lot, but we don't update the VERSION very often. By moving the "hard-coded" version out of the Makefile and setting it via an environment variable, we have to type a lot every time we build (`make build -j8` versus `make build -j8 VERSION=0.5.0`). To me, this seems like some extra operational overhead.

imxw:
> I have only one small concern: we run `make build` a lot, but we don't update the VERSION very often. By moving the "hard-coded" version out of the Makefile and setting it via an environment variable, we have to type a lot every time we build (`make build -j8` versus `make build -j8 VERSION=0.5.0`). To me, this seems like some extra operational overhead.
> 
> What do you think? @imxw @daniel-hutao

I forgot to emphasize that the main purpose here is that when developers build locally, the `dtm` version is not the default 0.4.0, but v0.4.0-41-g436bad5 to avoid confusion with the latest release version on github.

IronCore864:
> I forgot to emphasize that the main purpose here is that when developers build locally, the `dtm` version is not the default 0.4.0, but v0.4.0-41-g436bad5 to avoid confusion with the latest release version on github.

Got it. The purpose is to improve developers' experience when doing local development while keeping the original releasing process unchanged.

A very good proposal. Merging now. Thanks! @imxw

IronCore864 (approved):

PR #444, opened 2022-05-06
refactor: modify upload plugin from github to s3

KeHaohaoke:
# Summary
Upload plugin modification: from github to s3

## Key Points

## Description
Upload plugin modification: from GitHub to s3, and use CloudFlare as CDN. Use AWS official command line tools to make every uploaded file publicly readable.
Automatically create the corresponding version number directory, and the version number directory name is obtained from the tag of GitHub.

### Screenshots

- **AWS directory structure**

**Set up automatically in github actions. All files under this bucket can be accessed publicly, instead of being set in AWS console, otherwise it will cause security alarm.**

**In order to use the cdn of cloudflare, the bucket name must be set to the domain name form, otherwise AWS S3 will prompt that the bucket cannot be found when accessing.**

image

- **Upload to s3 through github actions succeeded.**

image

- **init** 

**Just for debugging and verification, the download address is printed out as the address of clouflare. This part of the code is not  submitted**

image

- **Terraform  Test : init**
    
    image

-  **Terraform  Test : plan**

image
image
image
image

-  **Terraform  Test : apply**

     **The error occurs because the bucket has been created manually, which is normal and in line with expectations.**

image
image
image
image

image

-  **test by AWS  policy simulator console**
    
image

-  **test cloudfront download link**

**response code is 200 OK**
    
image

-  **test direct s3 download link**
   
**response code is 403 forbbiden**
   
image


hack/terraform/config.tf
```diff
@@ -14,7 +14,16 @@ terraform {
...
+
+variable "s3_bucket_names" {
+  type = list(string)
+  default = ["download.devstream.io", "just-for-test"]
+}
+
+resource "aws_s3_bucket" "devstream_buckets" {
+  count         = length(var.s3_bucket_names)
+  bucket        = var.s3_bucket_names[count.index]
+}
```
IronCore864, line 29: No newline at the end of the file.

.github/workflows/automated-release.yml
```diff
@@ -72,6 +78,12 @@ jobs:
         run: make build -j8
       - name: Install Github-release
         run: go install github.com/github-release/github-release@latest
+      - name: Configure AWS credentials
```
IronCore864, line 81: We have configured aws credentials in previous steps. Do we need to reconfigure it again here?

hack/terraform/config.tf
```diff
@@ -14,7 +14,16 @@ terraform {
...
   }
 }
-
 provider "aws" {
   region = "ap-southeast-1"
 }
+
+variable "s3_bucket_names" {
+  type = list(string)
+  default = ["download.devstream.io", "just-for-test"]
```
IronCore864, line 23: Please remove "just-for-test"

hack/terraform/config.tf
```diff
@@ -14,7 +14,16 @@ terraform {
...
+}
+
+resource "aws_s3_bucket" "devstream_buckets" {
+  count         = length(var.s3_bucket_names)
```
IronCore864, line 27: Do not use "count". Please use "for_each" https://www.terraform.io/language/meta-arguments/for_each.

hack/terraform/devstream-iam.tf
```diff
@@ -129,6 +129,62 @@ resource "aws_iam_group_policy" "devstream-eks" {
...
+  name  = "DevStream-Download-Bucket-RW-Policy"
+  group = aws_iam_group.devstream.name
+
+  policy = jsonencode({
+    "Version": "2012-10-17",
+    "Statement": [
+      {
+        "Sid": "VisualEditor0",
+        "Effect": "Allow",
+        "Action": [
```
IronCore864, line 142: I do not think we need this much permission just to upload to/download from a bucket. Please re-test and ensure only the minimum required permission.

internal/pkg/pluginmanager/downloader_test.go
```diff
@@ -23,5 +23,5 @@ func TestDownload(t *testing.T) {
 func TestDownloadNotFound(t *testing.T) {
 	c := NewDownloadClient()
 	err := c.download(".", "doesntexist", "0.0.1")
-	assert.Contains(t, err.Error(), "404")
+	assert.Contains(t, err.Error(), "403")
```
IronCore864, line 26: Please add a comment here to explain why "not found" is 403 instead of 404 for an S3 bucket.

hack/terraform/config.tf
```diff
@@ -14,7 +14,16 @@ terraform {
...
 provider "aws" {
   region = "ap-southeast-1"
 }
+
+variable "s3_bucket_names" {
+  type = list(string)
+  default = ["download.devstream.io", "just-for-test"]
+}
+
+resource "aws_s3_bucket" "devstream_buckets" {
```
IronCore864, line 26: There are a few issues with this bucket:

The objects in the bucket are public.

For example, you can download this file https://s3.ap-southeast-1.amazonaws.com/download.devstream.io/v0.4.0/argocd-darwin-amd64_0.4.0.md5 directly with the S3 URL. 

This _should not_ be allowed.

If the user uses this URL, we will pay for the data transfer fees.

Solution: [S3 static website hosting](https://docs.aws.amazon.com/AmazonS3/latest/userguide/WebsiteHosting.html) with [CORS](https://docs.aws.amazon.com/AmazonS3/latest/userguide/cors.html). We should only allow users to download the content from download.devstream.io instead of using other domain names.

Make sure Cloudflare is enabled as well.

IronCore864 (changes requested):

IronCore864:
Please also delete all buckets/subdirectories for testing purposes.

hack/terraform/config.tf
KeHaohaoke: Count has been replaced by for_each.

hack/terraform/config.tf
KeHaohaoke: Newline have been added.

internal/pkg/pluginmanager/downloader_test.go
KeHaohaoke: For public users instead of the group we specified, it is not allowed to have the permission of list bucket, so for nonexistent object, 403 is returned instead of 404.

IronCore864: I think it's better to put what you said in the code as a line of comment. Because by default, developers will consider "NOT FOUND" as 404, not 403.

KeHaohaoke: Has been add comment in the code comments.

hack/terraform/devstream-iam.tf
KeHaohaoke: Limit the minimum available permissions in terraform

KeHaohaoke:
> Please also delete all buckets/subdirectories for testing purposes.

Unneeded directories have been deleted.

daniel-hutao (approved):

Issue #378, opened 2022-04-01
🍀 `Proposal`: Add Remote State Support


algobot76:
## Description

It is easy to use `dtm` for personal use. However, it's might not be enough for teams: everyone has a copy of state that may or may not be up-to-date and it's hard to manually update the copies to the most recent state as they are stored across different computers.  A remote seems like a solution to this problem.

## Describe the Proposal

To support remote state, `dtm` must satisfy the following requirements:

- Remote storage: `dtm` should be able to store the current state on a server (e.g. AWS S3).

- Locking: `dtm` should prevent concurrent writes to the remote state, otherwise, data corruption may occur.
- Security: `dtm` should not store the state in plain text. Encryption is needed to prevent security issues.
- Synchronization: `dtm` should provide a painless way for team members to get the latest state. Also, any instructions to update an old copy of the state should be not be executed.

## Has the Feature been Requested Before?

no

## Describe the Alternatives You Have Considered

n/a

## Additional Context

n/a

IronCore864:
A very good proposal!

At the moment, we do not have a whole lot of users yet, so maybe we can leave the state locking part out; an AWS S3 backend would be highly appreciated.

algobot76:
> A very good proposal!
> 
> At the moment, we do not have a whole lot of users yet, so maybe we can leave the state locking part out; an AWS S3 backend would be highly appreciated.

S3 backend will be a good starting point :)

daniel-hutao:
Since the tools we want to manage are mainly open source, why not support a free "remote backend" first? Such as ETCD/MySQL and so on. These backends will also be more readily available for many users. I also believe many users' environment is air-gapped, or whose teams do not have a chance to use commercial cloud storage.

If we implement ETCD backend first, we don't have to deploy ETCD on "remote host". Of course, ETCD can also be deployed on the machine where `dtm` is located. At this time, ETCD is essentially **local**, but we can call it **remote**, because when a user chooses to share an ETCD across teams, it is **remote**, and `dtm` does NOT Perceive how far this distance is.

IronCore864:
I think both s3 and etcd/mysql have their own values.

S3 has a great "pro", which is it's almost a universal protocol, supported by many cloud providers and even open-source databases. The con, of course, is the pricing (although minimum).

etcd is open-source, free to use, but the users will have to deploy it themselves. Similar for mysql.

If we have to choose one, which do you guys prefer?

daniel-hutao:
[...]

PR #446, opened 2022-05-06
feat: remote state config refactor

lfbdev:
# Summary

- refactor remote state
- new config file with remote state

## Key Points

## Description

- delete "--var-file" parameter
- command change to ./dtm apply -f config.yaml

- support configfile.yaml：
```
# var file is the ...
#varFile: gitops-variables.yaml # defaults to ./variables.yaml

# tool file is the ...
#  As the following, relative path, will join configfile.yaml dir and argocd.yaml  
toolFile: argocd.yaml  

#  As the following, if it is a absolute path, will return it directly
# toolFile: /Users/username/Projects/devstreamexamples/argocd.yaml

# state ...
state:
  backend: local
  options:
    stateFile: devstream.state
```

### Related Issues
#378

### Screenshots

image
### Other Information


internal/pkg/configloader/config.go
```diff
@@ -19,6 +21,19 @@ var (
 // Config is the struct for loading DevStream configuration YAML files.
 type Config struct {
 	Tools []Tool `yaml:"tools"`
+	State *State
+}
+
+// GeneralConfig is the struct for loading State and configuration YAML files.
+type GeneralConfig struct {
```
IronCore864, line 28: I suggest changing "GeneralConfig" to "ConfigFile" (only an example; the idea is to relate it to the actual config.) 

Otherwise, "Config" and "GeneralConfig" are very similar; developers can't tell their names and usage of them at a glance. "Config" is what the "LoadConf" function returns, and "GeneralConfig" is actually the config.yaml file.

Other comments need to be updated as well related to this.

IronCore864 (changes requested):

internal/pkg/configloader/config.go
daniel-hutao: +1 from me.
GeneralConfig has no usage here.

IronCore864 (approved):

Issue #466, opened 2022-05-11
💣 `Bug`: the process will crash when config file is not correct


aeinrw:
## Describe the Bug

https://github.com/devstream-io/devstream/blob/7e1c9a6a269ae76977f2ab34bf86d871174db3cb/internal/pkg/configloader/validation.go#L78-L98

The process will panic when `c.State` is nil because of nil pointer reference in `c.State.Options`. 

By the way, how to write a config file in v.0.5.0?

daniel-hutao:
Yeah, it should be:

```go
func validateConfigFile(c *ConfigFile) []error {
	errors := make([]error, 0)

	if c.ToolFile == "" {
		errors = append(errors, fmt.Errorf("tool file is empty"))
	}

	if c.State == nil {
		errors = append(errors, fmt.Errorf("state config is empty"))
		return errors
	}

	if c.State.Options == nil {
		errors = append(errors, fmt.Errorf("state options is empty"))
	}

	if c.State.Backend == "" {
		errors = append(errors, fmt.Errorf("backend is empty"))
	}
	return errors
}
```

Would please summit a pr to fix it? @aeinrw 

---

The config example can be seen here: `examples/quickstart.yaml` in release-0.5 branch. We'll merge it to main today.

daniel-hutao:
#467

daniel-hutao:
#468

IronCore864:
@lfbdev please double check if your PR from yesterday has already fixed this.

lfbdev:
It has been resolved, see: https://github.com/devstream-io/devstream/pull/473

IronCore864:
@aeinrw thanks for finding out about this critical issue.

I knew that you were trying to provide a fix for it, but due to internal miscommunication, this bug was also fixed by someone else here: https://github.com/devstream-io/devstream/pull/473.

Sorry again. This should've been your contribution. But hey, don't get disappointed. See you in the next PR :)

PR #473, opened 2022-05-12
fix: fix file validation

lfbdev:
# Summary
 fix file validation

- varfile not null validation
- toolfile not null validation
- state not null validation

### DIFF
--- a/internal/pkg/configloader/config.go
+++ b/internal/pkg/configloader/config.go
@@ -143,19 +143,15 @@ func LoadToolConf(toolFileName, varFileName string) *Config {
 func genToolVarPath(configFileName string, gConfig ConfigFile) (string, string, error) {
 	var absToolFilePath, absVarFilePath string
 	var err error
+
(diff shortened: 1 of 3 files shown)


internal/pkg/configloader/config.go
```diff
@@ -84,7 +84,7 @@ func LoadConf(configFileName string) (*Config, error) {
 	errs := validateConfigFile(&gConfig)
 	if len(errs) != 0 {
 		for _, e := range errs {
-			log.Errorf("General config validation failed: %s.", e)
+			log.Errorf("config file validation failed: %s.", e)
```
daniel-hutao, line 87: Config

internal/pkg/configloader/validation.go
```diff
@@ -85,14 +85,14 @@ func validateConfigFile(c *ConfigFile) []error {
 
 	if c.State == nil {
```
daniel-hutao, line 86: Don't change it, let community contributor do it later.

https://github.com/devstream-io/devstream/issues/466

IronCore864 (approved):

PR #494, opened 2022-05-17
📖`Docs`: update code references in core-concetps

aFlyBird0:
## Description
update code references in core-concetps

## ScreenShots

> One of the major part of the Config is a list of tools, defined in here
image

> Each Tool has its Name, Plugin, and Options, as defined here
image

> The dependency dependsOn is an array of strings, with each element being a dependency. Each dependency is named in the format of "NAME.PLUGIN".
image

> The State is actually a map of states, as defined here.
image

> Each state in the map is a struct containing Name, Plugin, Options, and Resource, as defined here.
image

### DIFF
--- a/docs/core-concepts/core-concepts.md
+++ b/docs/core-concepts/core-concepts.md
@@ -4,16 +4,16 @@ The architecture documentation explains how in general DevStream works. If you h
 
 ## 1 Tool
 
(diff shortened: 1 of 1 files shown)


aFlyBird0:
btw, should pr title use icon like 📖, or just docs?

daniel-hutao:
> btw, should pr title use icon like 📖, or just docs?

📖 `Docs`: xxx

daniel-hutao (approved):

Issue #502, opened 2022-05-17
:bomb: `Bug`: Maybe there are print errors in `/cmd/plugin`


aFlyBird0:
### What happened?

I'm not sure if the variables at the bottom of each file in `/cmd/plugin` are in the wrong order.

eg: 
```Go
const NAME = "argocd"
type Plugin string
var DevStreamPlugin Plugin

func main() {
	log.Infof("%T: %s is a plugin for DevStream. Use it with DevStream.\n", NAME, DevStreamPlugin)
}
```
I think it should be `DevStreamPlugin, NAME`.

Maybe I was wrong and I would be grateful if you tell me how it works.

### How to reproduce?

Just run `main()` func of any go file in `/cmd/pulgin` dir.

### DevStream version (to find out the version run: dtm version)

latest.

IronCore864:
@aFlyBird0 thanks for your attention to detail.

In fact, each go plugin shouldn't have a main() func at all; we only have it there to shut the IDE warnings up, LOL.

Would you care to contribute a PR fixing all of those occurrences?

Thanks!

aFlyBird0:
> @aFlyBird0 thanks for your attention to detail.
> 
> In fact, each go plugin shouldn't have a main() func at all; we only have it there to shut the IDE warnings up, LOL.
> 
> Would you care to contribute a PR fixing all of those occurrences?
> 
> Thanks!

Thank you! I'm willing to do that!

(Yep, go plugin do not need a main func. So I gotconfused. Thanks for your reply.)

PR #503, opened 2022-05-18
fix: fix go plugin log error

aFlyBird0:
## Description
fix go plugin log error

## Related Issues
#502

### DIFF
--- a/cmd/plugin/argocd/main.go
+++ b/cmd/plugin/argocd/main.go
@@ -35,5 +35,5 @@ func (p Plugin) Read(options map[string]interface{}) (map[string]interface{}, er
 var DevStreamPlugin Plugin
 
 func main() {
-	log.Infof("%T: %s is a plugin for DevStream. Use it with DevStream.\n", NAME, DevStreamPlugin)
+	log.Infof("%T: %s is a plugin for DevStream. Use it with DevStream.\n", DevStreamPlugin, NAME)
 }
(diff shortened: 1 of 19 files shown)


daniel-hutao:
I have a question about this change: how to test it?

image

@IronCore864 @aFlyBird0

daniel-hutao:
@IronCore864 Should we delete the `main()`? I deleted it and don't saw any errors.

Maybe we can refactor all `main.go` with plugins like below:

```go
package main

import (
	"github.com/devstream-io/devstream/internal/pkg/plugin/argocd"
)

// NAME is the name of this DevStream plugin.
const NAME = "argocd"

// Plugin is the type used by DevStream core. It's a string.
type Plugin string

// DevStreamPlugin is the exported variable used by the DevStream core.
var DevStreamPlugin Plugin

// Create implements the create of ArgoCD.
func (p Plugin) Create(options map[string]interface{}) (map[string]interface{}, error) {
	return argocd.Create(options)
}

// Update implements the update of ArgoCD.
func (p Plugin) Update(options map[string]interface{}) (map[string]interface{}, error) {
	return argocd.Update(options)
}

// Delete implements the delete of ArgoCD.
func (p Plugin) Delete(options map[string]interface{}) (bool, error) {
	return argocd.Delete(options)
}

// Read implements the read of ArgoCD.
func (p Plugin) Read(options map[string]interface{}) (map[string]interface{}, error) {
	return argocd.Read(options)
}
```

aFlyBird0:
> 

In fact, this main function will not be executed. But it's really wrong.

> from IronCore: we only have it there to shut the IDE warnings up

I guess it is used to avoid the waning of "unused var"?

IronCore864:
golangci-lint would report an error if we deleted the main() func.

Also, my IDE (VSCode) would give an error because of lacking the main() func.

I suggest that we keep the main() func so that beginners have less questions when they just get started.

daniel-hutao (approved):
/lgtm
/approve

Issue #511, opened 2022-05-18
:four_leaf_clover: `Proposal`: Using `go-playground/validator` to Validate Plugin Config Fields


iyear:
### What would you like to add? Why is this needed?

Currently for each field we manually validate, I think we can import a new dependency, https://github.com/go-playground/validator

we can just write tag in the fields of the struct, and the validation I proposed in #482 can be registered directly as tag to use.

daniel-hutao:
@iyear Cooooooool proposal! I agree to use this tool!

PR #533, opened 2022-05-18
refactor: use validator to validate fields

iyear:
#511 
I didn't test every plugin, but I checked them all as much as I could.

Some of the validation rules are too complicated. For example: `githubactions/golang/validate.go` , I kept some of the original validation rules.

### DIFF
--- a/internal/pkg/plugin/githubactions/golang/validate.go
+++ b/internal/pkg/plugin/githubactions/golang/validate.go
@@ -2,6 +2,7 @@ package golang
 
 import (
 	"fmt"
+
 	"github.com/devstream-io/devstream/pkg/util/validator"
 )
 
--- a/internal/pkg/plugin/githubactions/golang/validate.go
+++ b/internal/pkg/plugin/githubactions/golang/validate.go
@@ -2,47 +2,17 @@ package golang
 
 import (
 	"fmt"
+	"github.com/devstream-io/devstream/pkg/util/validator"
 )
 
 // validate validates the options provided by the core.
 func validate(opts *Options) []error {
 	retErrors := make([]error, 0)
 
-	// owner/org/repo/branch
-	if opts.Owner == "" && opts.Org == "" {
-		retErrors = append(retErrors, fmt.Errorf("owner and org are empty"))
+	if errs := validator.Struct(opts); len(errs) != 0 {
+		retErrors = append(retErrors, errs...)
 	}
-	if opts.Repo == "" {
-		retErrors = append(retErrors, fmt.Errorf("repo is empty"))
-	}
-	if opts.Branch == "" {
-		retErrors = append(retErrors, fmt.Errorf("branch is empty"))
-	}
-
-	// language
-	if opts.Language == nil {
-		retErrors = append(retErrors, fmt.Errorf("language is empty"))
-	}
-	if errs := opts.Language.Validate(); len(errs) != 0 {
-		for _, e := range errs {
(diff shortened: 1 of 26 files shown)


iyear:
Please review carefully, there are too many changes😥

internal/pkg/plugin/githubactions/golang/validate.go
```diff
@@ -2,43 +2,17 @@ package golang
 
 import (
 	"fmt"
+	"github.com/devstream-io/devstream/pkg/util/validator"
```
daniel-hutao, line 5: A blank line is required here

internal/pkg/plugin/jiragithub/options.go
```diff
@@ -2,11 +2,11 @@ package jiragithub
 
 // Options is the struct for configurations of the jiragithub plugin.
 type Options struct {
-	Owner          string
-	Org            string
-	Repo           string
+	Owner          string `validate:"required_without=Org"`
+	Org            string `validate:"required_without=Owner"`
+	Repo           string `validate:"required"`
 	JiraBaseUrl    string
```
daniel-hutao, line 8: JiraBaseUrl, JiraUserEmail, JiraProjectKey are required too.

pkg/util/validator/validator.go
```diff
@@ -0,0 +1,33 @@
+package validator
+
+import (
+	"errors"
```
daniel-hutao, line 4: fmt is needed, please use `make build` or `make fmt` locally.

daniel-hutao (changes requested):

IronCore864:
I've tested a normal use case for the following plugins:

- github-repo-scaffolding-golang
- githubactions-golang
- argocd
- argocdapp

And it worked.

Didn't regression-test all plugins or abnormal cases.

IronCore864:
@iyear thanks for this great improvement.

internal/pkg/plugin/trellogithub/workflow.go
```diff
@@ -18,10 +18,10 @@ var trelloWorkflow = &github.Workflow{
...
 type Options struct {
-	Owner       string
-	Org         string
-	Repo        string
-	Branch      string
+	Owner       string `validate:"required_without=Org"`
+	Org         string `validate:"required_without=Owner"`
+	Repo        string `validate:"required"`
+	Branch      string `validate:"required"`
 	BoardId     string
```
IronCore864, line 25: I think these three IDs are mandatory, too. This might have been a mistake earlier. Please also fix it.

IronCore864 (changes requested):

IronCore864:
A very good proposal and an elegant solution. Keep up the good work!

It's not easy to test all plugins in this PR. I'll create an issue for each plugin to improve.

@iyear

internal/pkg/plugin/trellogithub/workflow.go
iyear: ok,I will work for it.

iyear: Maybe take a look at what errors were there before and I will fix them. @IronCore864

iyear: and there are four IDs

IronCore864: Let's tackle this in a separate issue.

daniel-hutao (approved):
/LGTM
Great Job!

PR #539, opened 2022-05-19
Update cmd_apply.go

### LINKED ISSUE #537: :four_leaf_clover: `Proposal`:   Need a review on cmd_apply.go
###  Need a review

fmt.Errrof   & log.Errorf  is used in line 21 and 26 but NO string argument is passed. 

Screenshot from 2022-05-19 03-51-14

shubham-cmyk:
Update the log.Errorf to log.Error

## Related Issues

close #537

### DIFF
--- a/internal/pkg/pluginengine/cmd_apply.go
+++ b/internal/pkg/pluginengine/cmd_apply.go
@@ -23,7 +23,7 @@ func Apply(configFile string, continueDirectly bool) error {
 
 	err = pluginmanager.CheckLocalPlugins(cfg)
 	if err != nil {
-		log.Errorf("Error checking required plugins. Maybe you forgot to run \"dtm init\" first?")
+		log.Error("Error checking required plugins. Maybe you forgot to run \"dtm init\" first?")
 		return err
 	}
 


daniel-hutao:
Hi @shubham-cmyk , are you in our WeChat group? Or do you have my WeChat?

shubham-cmyk:
> Hi @shubham-cmyk , are you in our WeChat group? Or do you have my WeChat?

I am not in wechat group, also I can't join because I am not from China but I have already join the DevStream Slack channel.

IronCore864:
@shubham-cmyk would you please also change line 21 in the same file? Thanks!

internal/pkg/pluginengine/cmd_apply.go
```diff
@@ -23,7 +23,7 @@ func Apply(configFile string, continueDirectly bool) error {
 
 	err = pluginmanager.CheckLocalPlugins(cfg)
 	if err != nil {
-		log.Errorf("Error checking required plugins. Maybe you forgot to run \"dtm init\" first?")
+		log.Error("Error checking required plugins. Maybe you forgot to run \"dtm init\" first?")
```
IronCore864, line 26: Please also change line 21 for the same.

IronCore864 (changes requested):

internal/pkg/pluginengine/cmd_apply.go
shubham-cmyk: Since the fmt.Error doesn't exist so I have to change it either to log.Error or fmt.Print 
that might loose the lable of [Info] if I go for fmt.Print @daniel-hutao please into the same.

daniel-hutao: `fmt.Errorf` don't need to change for the `fmt.Error` doesn't exist. 

https://github.com/devstream-io/devstream/issues/537#issuecomment-1131061143

daniel-hutao (approved):

Issue #485, opened 2022-05-14
🍀 `Proposal`: S3 Backend


algobot76:
# Proposal: S3 Backend

## Description

Implement an S3 backend so that we can store the state file in S3.

## Assumptions

- Use `aws-sdk-go-v2` for interactions with AWS apis.
- Credentials stored in `~/.aws/credentials
- Use mocks for unit testing.

### S3Backend

```go
type S3Backend struct {
    file *S3File
}
```

```go
func NewS3Backend(bucket, region, key string) *S3Backend
```

```go
func (b *S3Backend) Read() (byte[], error)
```

```go
func (b *S3Backend) Write(data byte[]) error
```

### S3File

```go
type S3File struct {
    Bucket string
    Region string
    Key string
    ctx context.Context
    client *s3.Client
}
```

```go
func NewS3File(ctx context.Context, bucket, region, key string, client *s3.Client) (*S3File, error)
```

```go
func (f *S3File) Put(data []byte) error
```

```go
func (f *S3File) Download() ([]byte error)
```

## Has the Feature been Requested Before?

#378

IronCore864:
Related to https://github.com/devstream-io/devstream/issues/378.

PR #543, opened 2022-05-19
Refactor to support s3 backend

IronCore864:
## Description

- refactor to support different types of backends for state

## Related Issues

Related to https://github.com/devstream-io/devstream/issues/485 https://github.com/devstream-io/devstream/issues/378

N/A


internal/pkg/pluginengine/pluginengine_test.go
```diff
@@ -28,7 +28,7 @@ var _ = Describe("Pluginengine", func() {
 	BeforeEach(func() {
 		defer GinkgoRecover()
 
-		smgr, err = statemanager.NewManager()
+		smgr, err = statemanager.NewManager("local")
```
daniel-hutao, line 31: here should use const `Local` instead of "local"

IronCore864: This is handled inside `backend.GetBackend()`. The user of `statemanager` should not care too much about the implementation of `backend`.

daniel-hutao (approved):

Issue #505, opened 2022-05-18
:four_leaf_clover: `Proposal`: auto-release script improvement


daniel-hutao:
### What would you like to added? Why is this needed?

We have a release script: `build/package/auto-release-darwin-arm64.sh`, it starts with:

```sh
#! /bin/bash -e
set -o nounset

tag=v0.5.0
```

So we need to change this `tag` variable every time when we release a new version, which is obviously inelegant. So, why not set this `tag` with a parameter? Something like 

```sh
sh xxx.sh tag=v0.5.0
```

---

After the script improvement, please move the `auto-release-darwin-arm64.sh` and `upload_assets.sh` under `build/package` to `hack/release/` by the way.

yclchuxue:
I want to do this.

daniel-hutao:
@yclchuxue Go ahead and happy coding!

/assign @yclchuxue

PR #550, opened 2022-05-19
Feat: Auto-Release Script Improvement

yclchuxue:
## Pre-Checklist

- [ ] I have read through the [CONTRIBUTING.md](../CONTRIBUTING.md) documentation.

## Description

When a new version is released, the following methods can be used
```
sh xxx.sh tag=v0.5.0
```
## Related Issues

There will be a problem if more than one parameter is passed to the auto-release-darwin-arm64.sh file, and the first one is not tag=vx.x.x.

```
if [ $# -eq 1 ]
then
  info=$1
  tag=${info:4}
fi
```


summingyu:
Would it be better to use "getops"?@yclchuxue

yclchuxue:
Doesn't this require the input of parameters like "tag=v0.x.0"? If it is changed to "-t v0.x.0", it may not meet the requirements.Although I also feel it would be better to use getops.

yclchuxue:
@summingyu

daniel-hutao:
## Related Issues

close #505

summingyu:
Sorry, I checked the documentation, "getopts" does not support "long options", "getopt" support "long options" @yclchuxue

yclchuxue:
Ok,thanks.

Issue #557, opened 2022-05-20
:four_leaf_clover: `Proposal`: `dtm show status` Improvement


IronCore864:
### Background

Currently, `dtm` supports a `show status` subcommand:

```shell
tiexin@mbp ~/work/stream $ ./dtm show --help
Show is used to print some useful information.
Examples:
  dtm show config --plugin=A-PLUGIN-NAME
  dtm show status --plugin=A-PLUGIN-NAME --name=A-PLUGIN-INSTANCE-NAME
  dtm show status -p=A-PLUGIN-NAME -n=A-PLUGIN-INSTANCE-NAME 
```

As we can see, if we run `dtm show status -p=A-PLUGIN-NAME -n=A-PLUGIN-INSTANCE-NAME`, the "status" of that plugin and that instance will be printed.

### Improvement 1

We have changed our config file. Now the parameters for this command should be "plugin" and "instanceID", so "-p=A-PLUGIN-NAME -n=A-PLUGIN-INSTANCE-NAME" isn't precise any longer.

We suggest "-p=A-PLUGIN-NAME -i=INSTANCE-ID".

### Improvement 2

Would be nice if we can `dtm show status --all` or `dtm show status -a` so that the status of all instances of all plugins are printed out.

24sama:
Hi @IronCore864 @daniel-hutao 
Please assign this issue to me, I can try to fix it.

PR #562, opened 2022-05-20
fix: change flag "name" to "id" and add "all" flag

24sama:
## Pre-Checklist

## Description

* Change flag "name" to "id".
* Add "all" flag to `dtm show status` command.

## Related Issues

#557

### DIFF
--- a/cmd/devstream/show.go
+++ b/cmd/devstream/show.go
@@ -9,7 +9,8 @@ import (
 )
 
 var plugin string
-var instanceName string
+var instanceID string
+var statusAllFlag bool
 
 var showCMD = &cobra.Command{
 	Use:   "show",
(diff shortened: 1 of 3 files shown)


daniel-hutao:
@24sama Hello Leo Li, I tested this pr locally and found some error(Warn):

`./dtm show status --all`                                                                                                         

```sh
2022-05-20 14:19:29 ⚠ [WARN]  Empty instance name. Maybe you forgot to add --id=INSTANCE_ID. The default value "default" will be used.
```

Here may be some improvement can be done.

24sama:
> What if the user doesn't enter any flags? I think --all defaults to true will be more appropriate
```
dtm show status -p xxxx 
```
It will return all results, which isn't our expected.
I think we only need to add a little logic. If the user doesn't enter any flags, we make the value of `--all` to `true`.

daniel-hutao:
> Hi @daniel-hutao , thanks for your review.
> 
> > What if the user doesn't enter any flags? I think --all defaults to true will be more appropriate
> 
> ```
> dtm show status -p xxxx 
> ```
> 
> It will return all results if `--all` defaults to true, which isn't our expected. I think we only need to add a little logic. If the user doesn't enter any flags, we make the value of `--all` to `true`.
> 
> And I make some changes, PTAL.

Cool, your are right, the --all flag should be true only when no flags provided.

daniel-hutao (approved):
lgtm

PR #572, opened 2022-05-21
fix(configloader): varFile can be empty

aFlyBird0:
## Description

In the previous version, varFile was mandatory, but it should actually be optional.

### DIFF
--- a/cmd/devstream/list/list.go
+++ b/cmd/devstream/list/list.go
@@ -14,7 +14,7 @@ import (
 
 var PluginsName string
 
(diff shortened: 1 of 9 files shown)


aFlyBird0:
`LoadConf()` will be Called every time which contains `validateConfigFile()`.
bb86bceffb3d038b72ba451e33c93b8

Here are the core changes.

d9f099d79fddc2f709c5803e8bed599
image

daniel-hutao (approved):
LGTM
Great Job！！！

Issue #559, opened 2022-05-20
:four_leaf_clover: `Proposal`: Some Commands Do Not Need the "-f" Parameter


IronCore864:
### What would you like to add? Why is this needed?

At the moment, we have a global parameter "-f" which is required for commands that need access to the config file.

But not all commands need it, for example:
- `dtm show`
- `dtm list`
- `dtm develop`
- etc.

But when we run:

```shell
tiexin@mbp ~/work/stream $ ./dtm show --help
Show is used to print some useful information.
Examples:
  dtm show config --plugin=A-PLUGIN-NAME
  dtm show status --plugin=A-PLUGIN-NAME --name=A-PLUGIN-INSTANCE-NAME
  dtm show status -p=A-PLUGIN-NAME -n=A-PLUGIN-INSTANCE-NAME

Usage:
  dtm show [config | status] [flags]

Flags:
  -h, --help            help for show
  -n, --name string     specify name with the plugin instance
  -p, --plugin string   specify name with the plugin

Global Flags:
  -f, --config-file string   config file (default "config.yaml")
      --debug                debug level log
  -d, --plugin-dir string    plugins directory (default ".devstream")
  -y, --yes                  apply/delete directly without confirmation
```

The "-f" flag still shows up in the help for `dtm show` subcommand, even though `dtm show` won't need it.

We can change the "-f" flag to a non-global one so that it only shows up in necessary helps of certain subcommands.

yclchuxue:
I can do this

shubham-cmyk:
I would also like to give it a shot

daniel-hutao:
> I can do this

@yclchuxue It‘s a good first issue, let’s leave it to new contributor 😄

daniel-hutao:
> shot



> I would also like to give it a shot

Hi @shubham-cmyk, let’s leave this good first issue to the new contributor😁

shubham-cmyk:
@daniel-hutao but I have already solved it can you verify that #576 
I have removed the flag -f from the global and made specifically to the apply.go

daniel-hutao:
#576

@shubham-cmyk Please don't open a pr before claiming the relevant issue.

IronCore864:
Note that "show status" might need the "-f" param. Please test it thoroughly before submitting. Thanks!

himku:
I'll try this question.

daniel-hutao:
/assign @himku Happy coding!
Please test it thoroughly before submitting it. Thanks!

PR #576, opened 2022-05-21
Making -f Flag specifically to the dtm apply

shubham-cmyk:
@daniel-hutao  I have made the -f specific to the **dtm apply**

Would it be fine or should i go for local flag like : 
_localCmd.Flags().StringVarP(&Source, "source", "s", "", "Source directory to read from")_

## Related Issues
<!-- #559

### DIFF
--- a/cmd/devstream/apply.go
+++ b/cmd/devstream/apply.go
@@ -21,14 +21,6 @@ DevStream will generate and execute a new plan based on the config file and the 
 	Run: applyCMDFunc,
 }
 
-var applyConfigCMD = &cobra.Command{
-	Use:   "config",
-	Short: "Create or update DevOps tools according to DevStream configuration file",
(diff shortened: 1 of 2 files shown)


daniel-hutao:
Hi @shubham-cmyk, thanks for your pr!

But this implement is not correct, `-f` is not a flag exclusive to `apply`, for example `delete` is also needed; You have to be familiar with all subcommands of `dtm` and determine which subcommands need `-f` before you can start developing.

And please note that the commit message is clear, please read our [CONTRIBUTING.md](https://github.com/devstream-io/devstream/blob/main/CONTRIBUTING.md) first.

Please also pay attention to the specification of pr, we provide a template, please do not delete; in addition, please associate the corresponding issue in pr.

Finally, this is a good-first-issue, so let's leave it to new contributors.
Thanks again for the pr, but I had to close it.

Issue #566, opened 2022-05-20
:bomb: `Bug`: If Tool a Depends on Tool b, When b Fails to Install, a Will Still be Tried to Install


daniel-hutao:
### What Happened?

If tool a depends on tool b, when b fails to install, a will still be tried to install.

```sh
2022-05-20 15:06:14 ℹ [INFO]  Apply started.
2022-05-20 15:06:14 ℹ [INFO]  Using dir <.devstream> to store plugins.
2022-05-20 15:06:14 ℹ [INFO]  Using local backend: devstream.state.
2022-05-20 15:06:14 ℹ [INFO]  Tool (github-repo-scaffolding-golang/default) found in config but doesn't exist in the state, will be created.
2022-05-20 15:06:14 ℹ [INFO]  Tool (argocd/default) found in config but doesn't exist in the state, will be created.
2022-05-20 15:06:14 ℹ [INFO]  Tool (jira-github-integ/default) found in config but doesn't exist in the state, will be created.
2022-05-20 15:06:14 ℹ [INFO]  Tool (githubactions-golang/default) found in config but doesn't exist in the state, will be created.
2022-05-20 15:06:14 ℹ [INFO]  Tool (argocdapp/default) found in config but doesn't exist in the state, will be created.
2022-05-20 15:06:14 ℹ [INFO]  Start executing the plan.
2022-05-20 15:06:14 ℹ [INFO]  Changes count: 5.
2022-05-20 15:06:14 ℹ [INFO]  -------------------- [  Processing progress: 1/5.  ] --------------------
2022-05-20 15:06:14 ℹ [INFO]  Processing: (github-repo-scaffolding-golang/default) -> Create ...
2022-05-20 15:06:17 ℹ [INFO]  The repo dtm-test-go has been created.
2022-05-20 15:06:30 ✔ [SUCCESS]  Tool (github-repo-scaffolding-golang/default) Create done.
2022-05-20 15:06:30 ℹ [INFO]  -------------------- [  Processing progress: 2/5.  ] --------------------
2022-05-20 15:06:30 ℹ [INFO]  Processing: (argocd/default) -> Create ...
2022-05-20 15:06:30 ℹ [INFO]  -------------------- [  Processing progress: 3/5.  ] --------------------
2022-05-20 15:06:30 ℹ [INFO]  Processing: (jira-github-integ/default) -> Create ...
[...]

PR #580, opened 2022-05-22
feat(pluginmanager): a common plugin traverser && parallel download plugins

### LINKED ISSUE #579: :four_leaf_clover: `Proposal`: Enhancement: Parallel download plugins
## What Would You Like to Add? Why is This Needed?

Nowadays the plugins are downloaded serially, we can use parallel download to speed up.

## Design

Analyze the dependency graph and download them in batches, downloading only the plugins with dependency entry of 0 in each batch.

( Once any plugin download fails, end the entire download process)

## Anything else

Warning: There are many problems:

1. The download log may be messed up
2. The logic and syntax are complex and can easily lead to errors.
3. Maybe it is difficult to test.

aFlyBird0:
## Related Issues
close #579 

use `gitops.yaml` for example:

Here is the download patches test. (use plguins from `make build`)
USXG V(S9JET3{_ODWA0JV9

Let's change the name of `argocd` to test if download can be terminated automatically.
2X(C8C V)3FA2)GTC 062$U

## More Infomation.

1. **This is just a demo. I want to do this with everyone in DevStream**. 
2. Have fun finding bugs and refactoring it.
3. for structure: I dont't think the funcs in `DownloadPlugins` will be reuse. So I made them func vars. Maybe we can turn it into a parallel download class and convert all the function variables to methods of the class. This would be more conducive to testing.
4. It is difficult to test. Because now the dev version of the code is not downloading the plugin from the web but pre-compiling it. Maybe we can write another mock class for plugin downloads that looks the same as downloading a plugin from the web except that it reads the plugin locally.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,7 @@ require (
 	github.com/xanzy/go-gitlab v0.55.1
 	golang.org/x/crypto v0.0.0-20211215153901-e495a2d5b3d3
 	golang.org/x/oauth2 v0.0.0-20211104180415-d3ed0bb246c8
+	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 	gopkg.in/gookit/color.v1 v1.1.6
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 	helm.sh/helm/v3 v3.7.2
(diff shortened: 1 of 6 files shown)


aFlyBird0:
@daniel-hutao @IronCore864 

Please review, give me some ideas or dare to criticize, thanks!

aFlyBird0:
image

aFlyBird0:
@daniel-hutao 

Please see here, thanks very, very much. I have transformed it into a generic plugin traversal class (automatically analyzes plugin dependency order). You can choose to use sequential traversal or do parallel traversal for each batch.

If a failure is encountered during the traversal, it will stop all other traversals that depend on this plugin. This may be useful for #566.

All tests passed.

daniel-hutao:
I'll take a review now.

daniel-hutao:
I'm ashamed, this pr review up a bit difficult, I can not read. (perhaps very carefully line by line can be read, but the cost is too much); some feedback can give you: 

1. to solve specific problems, such as plug-in dependencies #566, so that we can know why this pr needs.
2. I can guess that there is a `dag` algorithm here, but `pluginmanager` does not need `dag`, perhaps it is only needed in `pluginengine`, you can look at the `topological_sort.go` in `pluginengine`
3. avoid too much indentation, especially large blocks of `for` loops (in fact, draw a small function/method out of the content of the `for`, so it will be much clearer. (Do not worry about the function is too simple, concise better reflect the tech level)

Issue #587, opened 2022-05-23
:four_leaf_clover: `Proposal`: Can GOPATH not need to be set?


linuxtmty:
Goimports often cannot be found, why not get the value of GOPATH directly in the Makefile, and let the user operate the environment variable。
It's a great experience to make a small change。
<img width="678" alt="image" src="https://user-images.githubusercontent.com/8682294/169755759-0e58a004-af24-4e83-b463-2ad7cbbfe521.png">
<img width="707" alt="image" src="https://user-images.githubusercontent.com/8682294/169755820-b94f242d-7aad-499a-92e6-76eda8882966.png">

iyear:
I think this makes sense.  I remember that GOPATH is not automatically added to PATH.

daniel-hutao:
@linuxtmty Thanks for the great proposal, you are right! Could you please help to create a pr to fix it?

iyear:
@linuxtmty  pr has not been created in the main repo.

PR #591, opened 2022-05-24
fix:gopath is no longer set

linuxtmty:
fix:gopath is no longer set


daniel-hutao:
close #587

@linuxtmty Thanks for the pr! This is a great optimization!

Also I hope you will pay attention to the format when submitting your pr. We provide a pr template, and while it's not strictly required to follow them, trying to keep them will make the information more readable.

PR #604, opened 2022-05-25
Feat: Config Logic Enhancement

daniel-hutao:

## Pre-Checklist

- [ ] I have read through the [CONTRIBUTING.md](../CONTRIBUTING.md) documentation.


aFlyBird0:
I think the structure could be made clearer further. @daniel-hutao 

For example, the current `renderVariables(varFileName string, configFileBytes []byte)` is fused based on `var filename` and `toolFilesConfigBytes`.

This results in forcing the specification of `varFile` to be standalone, which is hard to modify and expand.

Why don't we change the parameters to, `varFileMap` and `tooConfigBytes`? This would make it easy to generate `var map` in any way, such as from a separate file, or rather a small portion of an entire `config file`.

Even further, we could define the operations of getting `var map`, `tool bytes` as two interfaces. Then just implement a generic function that stitches together the final `config` from `var map` and `tool bytes`. Then write some implementation classes to read `var map` and `tool bytes` from a single file or multiple files.

Similarly, you can do the same for reading `state`.

Translated with www.DeepL.com/Translator (free version)

daniel-hutao:
> I think the structure could be made clearer further. @daniel-hutao
> 
> For example, `renderVariables(varFileName string, configFileBytes []byte)` is based on `var filename` and `toolFilesConfigBytes`.
> 
> This results in forcing the specification of `varFile` to be standalone, which is hard to modify and expand.
> 
> Why don't we change the parameters to, `varMap` and `toolConfigBytes`? This would make it easy to generate final `var map` in any way, such as from a separate file, or rather a small portion of an entire `config file`.
> 
> Even further, we could define the operations of getting `var map`, `tool bytes` as two interfaces. Then just implement a generic function that stitches together the final `config` from `var map` and `tool bytes`. Then write some implementation classes to read `var map` and `tool bytes` from a single file or multiple files.
> 
> Similarly, you can do the same for reading `state`.

@aFlyBird0 Please don't rush, it's still wip status; I'm just refactoring in steps, please wait until the final version to review

aFlyBird0:
Okay, just enjoy your refactor tour and notice me when finished. @daniel-hutao

daniel-hutao:
After more than an hour of git manipulation, I decided to abandon this pr and redevelop it based on the latest code because the wip time was too long and the mainline code had been changed too much and the conflict resolution was too costly.

PR #648, opened 2022-06-01
feat: add cf-ip-monitor

### LINKED ISSUE #560: :four_leaf_clover: `Proposal`: Cloudflare IP List Monitor
### Background

We use S3 to store published plugins from v0.6.0. And to speed things up, we have Cloudflare in front of the S3 bucket. The thing is, we grant access to S3 by IP addresses.

Cloudflare publishes its IP range here: https://www.cloudflare.com/ips/. It's also available from a text file: https://www.cloudflare.com/ips-v4.

If Cloudflare IP changes, we won't be able to `dtm init` anymore because S3 doesn't know if Cloudflare changes its IP or not and is still using the old list.

### IP Change Monitoring

It would be nice if we can have a daily cronjob that checks if the IP list of Cloudflare has changed or not.

If yes, send a notification (email/feishu/slack) to DevStream PMC.

### Functional Requirements

The script can be a simple python one, or a simple go app. We can even utilize github actions or something to that effect. The goal is to have a service like "hascloudflarechangedip.com" or something.

### Non-Functional Requirements

Maybe figure out a way to ensure 24 hours SLA.

I.E., at least run the job once per day so that we can be notified within 24 hours (maximum) after IP changed.

iyear:
## Pre-Checklist

## Description
Add Cloudflare IPv4 Monitor

1. Start workflow regularly each day to write `https://www.cloudflare.com/ips-v4` to the `ips-v4` file.
2. Use `git status` to check for diff.
3. If there is a diff, use the `github-actions bot` to commit and push, and create an issue(For details, see `ISSUE_TEMPLATE.md`)
4. If there is no diff, do nothing.

**`github-actions` is more difficult to test in the original repo, if there are problems after merging, I will continue to improve**

## Related Issues
close #560 

The screenshots only represent my test samples in the private repo, the files in this pr have been modified accordingly.

image
image


IronCore864 (approved):

IronCore864:
@iyear great solution. Merging now, let's give it a trail run.

Issue #642, opened 2022-05-31
:four_leaf_clover: `Proposal`:  dtm command auto-completion


prodanlabs:
## What Would You Like to Add? Why Is This Needed?

`dtm` command  auto-completion，make `dtm` easier to use

## Anything else

![image](https://user-images.githubusercontent.com/28917392/171205753-eadcaeb0-f648-45e4-b98d-91389f6d64b5.png)

![image](https://user-images.githubusercontent.com/28917392/171206182-27a01c66-d793-400e-8d9c-12eb75e680c2.png)

iyear:
There is currently no github bot to execute the command:(. And this issue looks similar to #627, but I don't know much about the related fields😐.

prodanlabs:
Oh, someone is doing it, right? Then I don't have to submit a PR.

aFlyBird0:
> There is currently no github bot to execute the command:(. And this issue looks similar to #627, but I don't know much about the related fields😐.

I think #627 is special for [`fig`](https://docs.devstream.io/en/latest/commands/autocomplete/#fig-autocomplete) , this issue is special for [`shell`](https://docs.devstream.io/en/latest/commands/autocomplete/#shell-autocomplete)

aFlyBird0:
Thank you very much for submitting so many issues for devstream!

However, there is already an official shell autocomplete (see [here](https://docs.devstream.io/en/latest/commands/autocomplete/#shell-autocomplete) for details), so some of the functions may be duplicated.

But what does the `-p` mean? Is it a regular filter?

aFlyBird0:
And to my partial understanding, `dtm show config -p` might be similar to `dtm list plugins -r`

![image](https://user-images.githubusercontent.com/36830265/171224843-98f2562b-8811-483e-a412-d433747490a7.png)

prodanlabs:
```
  -p, --plugin string   specify name with the plugin
```

prodanlabs:
```
$ ./dtm show config -p [TAB][TAB]
argocd                          githubactions-golang            github-repo-scaffolding-golang  gitlab-repo-scaffolding-golang  jenkins                         openldap
argocdapp                       githubactions-nodejs            gitlabci-generic                hashicorp-vault                 jira-github-integ               trello
devlake                         githubactions-python            gitlabci-golang                 helm-generic                    kube-prometheus                 trello-github-integ
$ ./dtm show config -p gitlab [TAB][TAB]
[...]

PR #650, opened 2022-06-01
feat: add command completion

prodanlabs:
Signed-off-by: prodan <pengshihaoren@gmail.com>

## Pre-Checklist

## Description

Auto-completion help, take `bash` as an example
```
$ ./dtm completion bash -h
generate autocompletion script for bash

Usage:
  dtm completion bash [flags]

Examples:
Load is completions in the current shell session:
# source <(dtm completion bash)

load completions for every new session:
(in Linux)# dtm completion bash > /etc/bash_completion.d/dtm
(in MacOS)# dtm completion bash > /usr/local/etc/bash_completion.d/dtm

Flags:
  -h, --help   help for bash

Global Flags:
      --debug   debug level log
```
1、`Flags` autocomplete
```bash
prodan@ubuntu:~/Desktop$ ./dtm show [TAB][TAB]
config  status  
prodan@ubuntu:~/Desktop$ ./dtm show config -[TAB][TAB]
--debug    -p         --plugin   --plugin=  
prodan@ubuntu:~/Desktop$ ./dtm show config -p [TAB][TAB] 
argocd                          githubactions-python            hashicorp-vault                 openldap
argocdapp                       github-repo-scaffolding-golang  helm-generic                    trello
devlake                         gitlabci-generic                jenkins                         trello-github-integ
githubactions-golang            gitlabci-golang                 jira-github-integ               
githubactions-nodejs            gitlab-repo-scaffolding-golang  kube-prometheus                 
prodan@ubuntu:~/Desktop$ ./dtm show config -p gitlab[TAB][TAB]
gitlabci-generic                gitlabci-golang                 gitlab-repo-scaffolding-golang  
prodan@ubuntu:~/Desktop$ ./dtm show config -p gitlab
```
2、`Args` autocomplete
```bash
prodan@ubuntu:~/Desktop$ ./dtm list plugins [TAB][TAB]
argocd                          githubactions-python            hashicorp-vault                 openldap
argocdapp                       github-repo-scaffolding-golang  helm-generic                    trello
devlake                         gitlabci-generic                jenkins                         trello-github-integ
githubactions-golang            gitlabci-golang                 jira-github-integ               
githubactions-nodejs            gitlab-repo-scaffolding-golang  kube-prometheus                 
prodan@ubuntu:~/Desktop$ ./dtm list plugins argocd[TAB][TAB]
argocd     argocdapp  
prodan@ubuntu:~/Desktop$ ./dtm list plugins argocd
```

3、Autocompletion when binary name is not `dtm`
```bash
prodan@ubuntu:~/Desktop$ source <(./dtm-linux-amd64 completion bash)
prodan@ubuntu:~/Desktop$ ./dtm-linux-amd64 [TAB][TAB]
apply       completion  delete      destroy     develop     help        init        list        show        verify      version     
prodan@ubuntu:~/Desktop$ ./dtm-linux-amd64 show [TAB][TAB]
config  status  
prodan@ubuntu:~/Desktop$ ./dtm-linux-amd64 show config -p [TAB][TAB]
argocd                          githubactions-python            hashicorp-vault                 openldap
argocdapp                       github-repo-scaffolding-golang  helm-generic                    trello
devlake                         gitlabci-generic                jenkins                         trello-github-integ
githubactions-golang            gitlabci-golang                 jira-github-integ               
githubactions-nodejs            gitlab-repo-scaffolding-golang  kube-prometheus                 
prodan@ubuntu:~/Desktop$ ./dtm-linux-amd64 show config -p 
```

## Related Issues

#642

### DIFF
--- a/cmd/devstream/completion.go
+++ b/cmd/devstream/completion.go
@@ -1,0 +1,61 @@
+package main
+
+import (
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+
+	"github.com/devstream-io/devstream/internal/pkg/completion"
+)
+
+func completionCMD(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                   "completion",
+		Short:                 "Generate the autocompletion script for dtm for the specified shell",
+		Long:                  "See each sub-command's help for details on how to use the generated script.",
+		DisableFlagsInUseLine: true,
+		Args:                  cobra.ExactValidArgs(1),
+	}
+
+	binaryName := filepath.Base(os.Args[0])
+	bash := &cobra.Command{
+		Use:     "bash",
+		Short:   "generate autocompletion script for bash",
+		Example: completion.BashExample(binaryName),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return completion.CompletionBash(out, cmd)
+		},
+	}
+
+	zsh := &cobra.Command{
+		Use:     "zsh",
+		Short:   "generate autocompletion script for zsh",
+		Example: completion.ZshExample(binaryName),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return completion.CompletionZsh(out, cmd)
+		},
+	}
+
+	fish := &cobra.Command{
+		Use:     "fish",
+		Short:   "generate autocompletion script for fish",
+		Example: completion.FishExample(binaryName),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return cmd.Root().GenFishCompletion(out, true)
+		},
+	}
+
+	powershell := &cobra.Command{
+		Use:     "powershell",
+		Short:   "generate autocompletion script for powershell",
+		Example: completion.PowershellExample(binaryName),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return cmd.Root().GenPowerShellCompletionWithDesc(out)
+		},
+	}
+	cmd.AddCommand(bash, zsh, fish, powershell)
+
+	return cmd
+}
--- a/internal/pkg/completion/completion.go
+++ b/internal/pkg/completion/completion.go
@@ -1,0 +1,100 @@
+package completion
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+
+	"github.com/devstream-io/devstream/cmd/devstream/list"
+	"github.com/devstream-io/devstream/pkg/util/log"
+)
+
+func FlagPluginsCompletion(cmd *cobra.Command, flag string) {
+	if err := cmd.RegisterFlagCompletionFunc(flag, func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return list.PluginsNameSlice(), cobra.ShellCompDirectiveDefault
+	}); err != nil {
+		log.Warn(err)
+	}
+}
+
+func FlagConfigFileCompletion(cmd *cobra.Command) {
+	validConfigFilenames := []string{"yaml"}
+
+	if err := cmd.Flags().SetAnnotation("config-file", cobra.BashCompFilenameExt, validConfigFilenames); err != nil {
+		log.Warn(err)
+	}
+}
+
+func CompletionBash(out io.Writer, cmd *cobra.Command) error {
+	err := cmd.Root().GenBashCompletion(out)
+
+	// The default binary name downloaded from the Releases page is dtm-{os}-amd64
+	// solve the problem that autocompletion fails when the name of the binary is not dtm
+	if binary := filepath.Base(os.Args[0]); binary != "dtm" {
+		renamedBinary := `
+# the user renamed the dtm binary
+if [[ $(type -t compopt) = "builtin" ]]; then
+    complete -o default -F __start_dtm %[1]s
+else
+    complete -o default -o nospace -F __start_dtm %[1]s
+fi
+`
+		fmt.Fprintf(out, renamedBinary, binary)
+	}
+
+	return err
+}
+
+func CompletionZsh(out io.Writer, cmd *cobra.Command) error {
+	err := cmd.Root().GenZshCompletionNoDesc(out)
+
+	// The default binary name downloaded from the Releases page is dtm-{os}-amd64
+	// solve the problem that autocompletion fails when the name of the binary is not dtm
+	if binary := filepath.Base(os.Args[0]); binary != "dtm" {
+		renamedBinary := `
+# the user renamed the dtm binary
+compdef _dtm %[1]s
+`
+		fmt.Fprintf(out, renamedBinary, binary)
+	}
+
+	fmt.Fprintf(out, "compdef _dtm dtm")
+
+	return err
+}
+
+func BashExample(binary string) string {
+	return fmt.Sprintf(`Load is completions in the current shell session:
+# source <(%s completion bash)
+
+Load completions for every new session:
+(in Linux)# %s completion bash > /etc/bash_completion.d/dtm
+(in MacOS)# %s completion bash > $(brew --prefix)/etc/bash_completion.d/dtm`, binary, binary, binary)
+}
+
+func ZshExample(binary string) string {
+	return fmt.Sprintf(`Load is completions in the current shell session:
+# source <(%s completion zsh)
+
+Load completions for every new session:
+# %s completion zsh > "${fpath[1]}/_dtm"`, binary, binary)
+}
+
+func FishExample(binary string) string {
+	return fmt.Sprintf(`Load is completions in the current shell session:
+#  %s completion fish | source
+
+Load completions for every new session:
+# %s completion fish > ~/.config/fish/completions/dtm.fish`, binary, binary)
+}
+
+func PowershellExample(binary string) string {
+	return fmt.Sprintf(`Load is completions in the current shell session:
+C:\> %s completion powershell | Out-String | Invoke-Expression
+
+Load completions for every new session:
+add the output of the above command to powershell profile.`, binary)
+}
--- a/cmd/devstream/apply.go
+++ b/cmd/devstream/apply.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/devstream-io/devstream/internal/pkg/completion"
 	"github.com/devstream-io/devstream/internal/pkg/pluginengine"
 	"github.com/devstream-io/devstream/pkg/util/log"
 )
@@ -12,7 +13,7 @@ import (
 var applyCMD = &cobra.Command{
 	Use:   "apply",
 	Short: "Create or update DevOps tools according to DevStream configuration file",
-	Long: `Create or update DevOps tools according to DevStream configuration file. 
+	Long: `Create or update DevOps tools according to DevStream configuration file.
 DevStream will generate and execute a new plan based on the config file and the state file by default.`,
 	Run: applyCMDFunc,
 }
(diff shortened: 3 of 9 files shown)


daniel-hutao:
@imxw Would you please help to take a review with this pr too?

imxw:
> @imxw Would you please help to take a review at this pr too?

Of course, it's my pleasure.

prodanlabs:
cc /daniel-hutao

cmd/devstream/completion.go
```diff
@@ -0,0 +1,132 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return cmd.Root().GenPowerShellCompletionWithDesc(out)
+		},
+	}
+	cmd.AddCommand(bash, zsh, fish, powershell)
+
+	return cmd
+}
+
+func completionBash(out io.Writer, cmd *cobra.Command) error {
```
IronCore864, line 63: We follow the [Standard Go Project Layout](https://github.com/golang-standards/project-layout), and the [`cmd`](https://docs.devstream.io/en/latest/development/project-layout/#cmd) directory is better not to contain a lot of logic:

> Don't put a lot of code in the application directory. If you think the code can be imported and used in other projects, then it should live in the /pkg directory. If the code is not reusable or if you don't want others to reuse it, put that code in the /internal directory. You'll be surprised what others will do, so be explicit about your intentions!

Based on this, I suggest we move the code from line 63 to the end of this file to another place, like `internal/pkg/completion/xxx.go` (only an example/a suggestion.)

cmd/devstream/completion.go
```diff
@@ -0,0 +1,132 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+)
+
+// completionCMD represents the completion command
```
IronCore864, line 12: This seems to be a little bit redundant information. I think it's fine if we don't have this comment at all.

IronCore864 (changes requested):

IronCore864:
@prodanlabs thanks for this great feature. I reviewed and added two little comments.

IronCore864:
After some testing, there is another comment I would like to add:

On macOS, given how bash-completion is installed, the default location of completion scripts isn't necessarily always `/usr/local/etc/bash_completion.d/dtm`.

For example, if auto-completion@2 is installed with brew, it seems the default location is `/opt/homebrew/etc/bash_completion.d`.

So, it's not very accurate to put a default location (especially for macOS, considering how many people are using brew) in the help doc. Maybe we can give some extra information in the help doc.

Another issue is, there are two versions of bash-completion, v1 and v2. V1 is for Bash 3.2 (which is the default on macOS), and v2 is for Bash 4.1+. I didn't test if the dtm completion script works with Bash 3.2 or not. I think it's best to tell macOS users how to upgrade to Bash 4.1+ (maybe brew install) and to install auto-completion@2.

K8s has a decent doc on auto-completion: https://kubernetes.io/docs/tasks/tools/included/optional-kubectl-configs-bash-mac/

We could also write a blog about it. Are you interested? @prodanlabs

prodanlabs:
Changes have been made as suggested by the comments, but I haven't seen this style `cobra` project.

prodanlabs:
> After some testing, there is another comment I would like to add:
> 
> On macOS, given how bash-completion is installed, the default location of completion scripts isn't necessarily always `/usr/local/etc/bash_completion.d/dtm`.
> 
> For example, if auto-completion@2 is installed with brew, it seems the default location is `/opt/homebrew/etc/bash_completion.d`.
> 
> So, it's not accurate to put a default location (especially for macOS, considering how many people are using brew) in the help doc. Maybe we can give some extra information in the help doc.
> 
> Another issue is, there are two versions of bash-completion, v1 and v2. V1 is for Bash 3.2 (which is the default on macOS), and v2 is for Bash 4.1+. I didn't test if the dtm completion script works with Bash 3.2 or not. I think it's best to tell macOS users how to upgrade to Bash 4.1+ (maybe brew install) and to install auto-completion@2.
> 
> K8s has a decent doc on auto-completion: https://kubernetes.io/docs/tasks/tools/included/optional-kubectl-configs-bash-mac/
> 
> We could also write a blog about it. Are you interested? @prodanlabs



> 


Personally, I think `Example` should be generic, as for other cases, I think you are right, you should write a document description. But unfortunately I can't do it because I don't have `MacOS`

cmd/devstream/list.go
```diff
@@ -21,7 +23,26 @@ var listPluginsCMD = &cobra.Command{
...
+
+func compListPlugins(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	if len(args) == 0 {
+		return list.PluginsNameSlice(), cobra.ShellCompDirectiveNoFileComp
+	}
+
+	if len(args) == 1 {
+		var choices []string
+		for _, pluginName := range list.PluginsNameSlice() {
+			if strings.HasPrefix(pluginName, args[0]) {
```
imxw, line 38: The purpose of `dtm list plugins` is to show all available plugins, so why filter the specified plugins here?

imxw:
@prodanlabs Thanks for your contribution, I like your idea! After testing, I have left a few comments. In addition, I wonder if you'd be interested in implementing this completion feature, where [TAB][TAB] only shows yaml type files when using `-f` flag.

internal/pkg/completion/completion.go
```diff
@@ -0,0 +1,81 @@
...
+	return err
+}
+
+func BashExample(binary string) string {
+	return fmt.Sprintf(`Load is completions in the current shell session:
+# source <(%s completion bash)
+
+Load completions for every new session:
+(in Linux)# %s completion bash > /etc/bash_completion.d/dtm
+(in MacOS)# %s completion bash > /usr/local/etc/bash_completion.d/dtm`, binary, binary, binary)
```
imxw, line 56: `/usr/local/` can be changed to `$(brew --prefix)`, because only Mac computers with Intel processors is `/usr/local/`, but M1 is `/opt/homebrew`

cmd/devstream/list.go
prodanlabs: same as `dtm list plugins -r`, filter plugin by regex.

imxw: Got it!

prodanlabs: Yes, do [TAB][TAB] to match or complete. complete is always like this.

imxw: > Yes, do [TAB][TAB] to match or complete. complete is always like this.

Yeah, but plugins here are not args or subcommands. It's just like `kubectl plugin list`.

prodanlabs: I don't think this is the problem. When the user wants to know how many plugins are related to `argocd`, they can simply do `dtm list plugins argocd[TAB][TAB]` to filter them out. I believe this is also the purpose of `dtm list plugins -r`.

prodanlabs:
> @prodanlabs Thanks for your contribution, I like your idea! After testing, I have left a few comments. In addition, I wonder if you'd be interested in implementing this completion feature, where [TAB][TAB] only shows yaml type files when using `-f` flag.


This doesn't make sense, neither `kubectl/helm` does filtering based on file suffix or name. So I will not change nor recommend changing it, to be honest.

imxw:
> 



> 



> > @prodanlabs Thanks for your contribution, I like your idea! After testing, I have left a few comments. In addition, I wonder if you'd be interested in implementing this completion feature, where [TAB][TAB] only shows yaml type files when using `-f` flag.
> 
> I don't think it makes much sense, to be honest.

It's okay, it's not mandatory to achieve.

internal/pkg/completion/completion.go
```diff
@@ -0,0 +1,119 @@
...
+
+func FlagPluginsCompletion(cmd *cobra.Command, flag string) {
+	if err := cmd.RegisterFlagCompletionFunc(flag, func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return list.PluginsNameSlice(), cobra.ShellCompDirectiveDefault
+	}); err != nil {
+		log.Warn(err)
+	}
+}
+
+func FlagConfigFileCompletion(cmd *cobra.Command) {
```
imxw, line 42: As I said before, it's not mandatory to achieve and you also didn't think it is necessary to achieve, but since you have achieved it, I refer to the official documentation and found that this implementation is not applicable to `Fish` and `PowerShell`

[Fish doesn't support `BashCompFilenameExt` ](https://github.com/spf13/cobra/blob/d8184d32696bee36f682bcb3ace5642af54cd7ad/shell_completions.md#limitations-1)

[Powershell doesn't support `BashCompFilenameExt`](https://github.com/spf13/cobra/blob/d8184d32696bee36f682bcb3ace5642af54cd7ad/shell_completions.md#limitations-2)

A more generic implementation is as follows.

https://github.com/spf13/cobra/blob/d8184d32696bee36f682bcb3ace5642af54cd7ad/shell_completions.md#specify-valid-filename-extensions-for-flags-that-take-a-filename

Of course, if you still don't think it's necessary to implement this feature, you can just remove the relevant code. Thanks again.

imxw (changes requested):

internal/pkg/completion/completion.go
```diff
@@ -0,0 +1,119 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/devstream-io/devstream/cmd/devstream/list"
+	"github.com/devstream-io/devstream/pkg/util/log"
+)
+
+func ArgsPluginsCompletion(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
imxw, line 16: > I don't think this is the problem. When the user wants to know how many plugins are related to `argocd`, they can simply do `dtm list plugins argocd[TAB][TAB]` to filter them out. I believe this is also the purpose of `dtm list plugins -r`.

I agree that we need to allow users to filter plugins. So `dtm list plugins -r` should be accepted. However, in my opinion, the premise `dtm list plugins argocd[TAB][TAB]` can be accepted is that `dtm list plugins` accepts the plugin name as arg or subcommand. I don't know if you understand what I mean.

 I think `dtm list plugins -r xxx` is good. The result can be used as `stdin` for other commands. For example
```
dtm list plugins -r "\bargocd\b" | dtm show config
dtm show config $(dtm list plugins -r "\bargocd\b")
```

The results of `dtm list plugins argocd`  are the same as `dtm list plugins`.  In addition, we have to clean up the command manually before executing the next command.

In summary, why do we need `dtm list plugins argocd[TAB][TAB]`. 

What do you think? @daniel-hutao @IronCore864

imxw (changes requested):

cmd/devstream/list.go
IronCore864: @prodanlabs the only ways to use the `list plugins` subcommand are:
- `dtm list plugins` which shows all the plugins
- `dtm list plugins -r argo` which shows the plugins whose names match "argo"

So, even if the user types `dtm list plugins argo [tab][tab]` and it shows `dtm list plugins argocd`, it's not useful, because the user can't execute this command.

I think that's why @imxw thought [this](https://github.com/devstream-io/devstream/pull/650/files#diff-3857dee2ec9cf176ffd4d69f926a87374ed4c37ce8a5db3ce5ac6ce8b737daa0R25) isn't necessary.

imxw:
The current zsh completion generated by cobra does not work, but the problem has been solved in [this pr](https://github.com/spf13/cobra/pull/1690), we can fix it by `go get github.com/spf13/cobra@master` or wait for cobra's version 1.5.0 @prodanlabs @daniel-hutao @IronCore864

internal/pkg/completion/completion.go
daniel-hutao: Right, the `dtm list plugins argocd` command does not exist; `dtm list plugins` is used to print all plugins, and this command has no further subcommands, it only has flags, like `-r`. So `dtm list plugins [tag]` should not output specific plugin names.

imxw (approved):

daniel-hutao (approved):

daniel-hutao (approved):

daniel-hutao:
> The current zsh completion generated by cobra does not work, but the problem has been solved in [this pr](https://github.com/spf13/cobra/pull/1690), we can fix it by `go get github.com/spf13/cobra@master` or wait for cobra's version 1.5.0 @prodanlabs @daniel-hutao @IronCore864

Please continue to follow up on this issue, I suggest using cobra stable version, please wait for the `1.5.0` version release to fix it.

PR #666, opened 2022-06-10
test(downloader): add downloader test case

summingyu:
## Description
1. improve coverage
2. Optimize return time in case of errors

## Current Behavior
image

image

image

image

### DIFF
--- a/pkg/util/downloader/downloader.go
+++ b/pkg/util/downloader/downloader.go
@@ -14,11 +14,6 @@ import (
 func Download(url, filename, targetDir string) (int64, error) {
 	log.Debugf("Target dir: %s.", targetDir)
 	log.Debugf("URL: %s.", url)
-	resp, err := http.Get(url)
-	if err != nil {
-		return 0, err
-	}
-
 	var fileName = filename
 	if fileName == "" {
 		fileName = filepath.Base(url)
@@ -28,12 +23,16 @@ func Download(url, filename, targetDir string) (int64, error) {
 		return 0, fmt.Errorf("failed to get the filename from url: %s", url)
 	}
 
-	err = os.MkdirAll(targetDir, 0755)
-	if err != nil {
+	if err := os.MkdirAll(targetDir, 0755); err != nil {
 		return 0, err
 	}
 
 	f, err := os.Create(filepath.Join(targetDir, fileName))
+	if err != nil {
+		return 0, err
+	}
+
+	resp, err := http.Get(url)
 	if err != nil {
 		return 0, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/util/downloader/downloader.go
```diff
@@ -14,11 +14,6 @@ import (
 func Download(url, filename, targetDir string) (int64, error) {
 	log.Debugf("Target dir: %s.", targetDir)
 	log.Debugf("URL: %s.", url)
-	resp, err := http.Get(url)
-	if err != nil {
-		return 0, err
-	}
-
 	var fileName = filename
```
IronCore864, line 17: It seems this isn't necessary.

How about:

```go
if filename == "" {
    filename = filepath.Base(url)
}
```
?

pkg/util/downloader/downloader.go
```diff
@@ -38,5 +32,10 @@ func Download(url, filename, targetDir string) (int64, error) {
 		return 0, err
 	}
 
+	resp, err := http.Get(url)
```
IronCore864, line 35: Very good adjustment. No need to download the file first if we can't even get the filename from the URL.

pkg/util/downloader/downloader_test.go
```diff
@@ -13,6 +13,13 @@ var _ = Describe("Downloader", func() {
 	Context("Downloader test", func() {
 		var url = "https://github.com/devstream-io/devstream/releases/download/v0.0.1/argocdapp_0.0.1.so"
 		var targetDir = "tmp"
+
+		It("Should return err", func() {
```
IronCore864, line 17: We can improve the name, for example:

It should return error if filename is empty and can't be parsed from the url.

IronCore864 (changes requested):

summingyu:
I think you're right, I'll change it now

aFlyBird0 (approved):
This helps to improve the speed of testing and increase test coverage, great!

pkg/util/downloader/downloader_test.go
aFlyBird0: Please pay attention here. @summingyu

pkg/util/downloader/downloader_test.go
```diff
@@ -13,6 +13,19 @@ var _ = Describe("Downloader", func() {
...
+			size, err := downloader.Download(url, ".", targetDir)
+			Expect(err).To(HaveOccurred())
+			Expect(size).To(Equal(int64(0)))
+		})
+
+		It("when url and filename are empty, Shold return err", func() {
```
aFlyBird0, line 23: *When    *should

aFlyBird0 (changes requested):

IronCore864 (approved):

IronCore864:
Thanks @summingyu for contributing more tests, and refactoring the existing logic!

I updated some of the descriptions of the test cases directly to fix grammar issues; hope you don't mind.

@aFlyBird0's comment is also fixed. Merging now. Due to duplicated commit messages, I'm squashing instead of merging. @summingyu please pay closer attention in the future. All in all, well done!

PR #821, opened 2022-07-06
Fix duplicate downloads and refactor code

### LINKED ISSUE #818: :bomb: `Bug`: Duplicate downloads
### What Happened?

When using the `init` command, `dtm` checks for non-existent plugins and pulls them from the remote. However, I find that every time I use `init`, it triggers an md5 mismatch and re-downloads, even though the corresponding ones have all downloaded successfully.

I found that 

https://github.com/devstream-io/devstream/blob/1d2c60493cc5327849862dce7e8f075591b5f3ba/internal/pkg/pluginmanager/manager.go#L75-L78

There is an error in the logic of this section. Here the download has already finished, but the `StatError` used is still the one before the download, which will cause `pluginFileErr == nil && pluginMD5FileErr == nil` to never hold.

I add some debug outputs.

`log.Debug(isMD5Match, pluginFileErr, pluginMD5FileErr)`

image

This means that every time a user uses `init`, twice as much traffic is consumed.

### How to Reproduce?

`dtm init -f quickstart.yaml --debug`

### Anything else

I will change the logic of this function and try to fix it.

Please assign me.

### DevStream Version

latest

iyear:
## Pre-Checklist

## Description
1. fix duplicate downloads logic
2. make outputs more beautiful
3. refactor some code

## Related Issues
close #818 

The version in 3 screenshots is a forced change during testing, and the commits have been submitted in accordance with the normal.

### Download all plugins from remote
full

### Download some plugins from remote
part

### No need to download plugins from remote
none

### Init in dev version
image

### DIFF
--- a/cmd/devstream/init.go
+++ b/cmd/devstream/init.go
@@ -28,7 +28,7 @@ func initCMDFunc(_ *cobra.Command, _ []string) {
 	}
 
 	if version.Dev {
-		log.Errorf("Dev version plugins can't be downloaded from the remote plugin repo; please run `make build-plugin.PLUGIN_NAME` to build it locally.")
+		log.Errorf("Dev version plugins can't be downloaded from the remote plugin repo; please run `make build-plugin.PLUGIN_NAME` to build them locally.")
 		return
 	}
 
--- a/cmd/devstream/init.go
+++ b/cmd/devstream/init.go
@@ -1,12 +1,15 @@
 package main
 
 import (
+	"fmt"
+
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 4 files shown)


IronCore864 (approved):

IronCore864:
Very good refactor. Well done! @iyear

IronCore864:
@hxcGit need another review here. Please have a local test if possible.

daniel-hutao (approved):
Good job!

aFlyBird0:
It's very cool! @iyear 🚀

PR #829, opened 2022-07-08
chore: refactor gitlab-ce-docker and fix some docs format

aFlyBird0:

## Pre-Checklist

## Description
1. some refactor in gitlab-ce-docker
2. fix the check of deleting volumes
3. show access URL after installing gitlab-ce-docker
4. update docs

## Anything else
Already tested locally


docs/contributor_ladder.zh.md
```diff
@@ -16,22 +16,22 @@ DevStream 项目的成员会尽可能的帮助所有的贡献者在贡献者阶
 _定义：直接为项目做出有价值的贡献。其表现形式可以不只是代码。非代码的贡献同样重要且被我们所认可！在 Contributor 级别的贡献者可能是一个新的 Contributor，或者只是偶尔提交一些贡献。_
 
 - 责任：
-  - 遵循 [CNCF CoC](https://github.com/cncf/foundation/blob/main/code-of-conduct.md)
-  - 遵循 [项目贡献指南](https://github.com/devstream-io/devstream/blob/main/docs/contributing_guide.zh.md)
+    - 遵循 [CNCF CoC](https://github.com/cncf/foundation/blob/main/code-of-conduct.md)
+    - 遵循 [项目贡献指南](https://github.com/devstream-io/devstream/blob/main/docs/contributing_guide.zh.md)
 - 要求（以下的一项或者多项 PS：多多益善 💖）：
```
daniel-hutao, line 21: > (信末签名后的）再者，又及；附言（略作P.s.,p.s.,P.S.,ps.,PS.,PS:)

我不确定规范写法是什么，比如 P.s. xxx

另外 PS 前是否需要逗号或者句号？

aFlyBird0: 中文资料说什么写都行，wiki百科没强调格式，但是用的是 "P.S."

image

至于这个逗号，我也拿捏不准，或许加一个更好，毕竟不能连读

algobot76:
Is it necessary to remove `deletion_test.go`?

internal/pkg/plugin/gitlabcedocker/create.go
```diff
@@ -41,11 +42,15 @@ func Create(options map[string]interface{}) (map[string]interface{}, error) {
...
 	}
 
+	// 4. check if the volume is created successfully
 	volumes, err := op.ListContainerMounts(gitlabContainerName)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get container mounts: %v", err)
 	}
 
+	// 5. show the access url
+	showGitLabUrl(opts)
```
algobot76, line 52: To be consistent with the rest of the codebase, it should be `showGitLabURL` instead.

algobot76 (changes requested):

internal/pkg/plugin/gitlabcedocker/create.go
aFlyBird0: > Is it necessary to remove `deletion_test.go`?

I think `deletion_test.go` doesn't work properly now.

aFlyBird0: > To be consistent with the rest of the codebase, it should be `showGitLabURL` instead.

Thank you!

algobot76: > > Is it necessary to remove `deletion_test.go`?
> 
> I think `deletion_test.go` doesn't work properly now.

If you think it's not working, why not improve it instead?:relaxed: A well-designed test case can serve as extra documentation for the codebase.

docs/contributor_ladder.zh.md
IronCore864: The [Cambridge Dictionary](https://dictionary.cambridge.org/us/dictionary/english/ps) suggests that PS is the proper format in British English.

Example: PS Don’t forget to let the cat in before you go to bed.

The Cambridge Dictionary also says that P.S. (with periods after each letter) is the American English format.

PS (postscript) comes from Latin "post scriptum," and for Latin abbreviations, periods is generally included. E.g.: i.e., Ph.D., a.m., p.m., Q.E.D.

So, while both PS and P.S. are correct, let's stick to the American English way, for two reasons:
- it follows the convention
- more people speaks American English than British English.

internal/pkg/plugin/gitlabcedocker/create.go
aFlyBird0: Thank you very much, I agree. The test I removed was actually only e2e test for most of the content. But inspired by you, I added as much as I could to the wish test, thanks!

internal/pkg/plugin/gitlabcedocker/delete_test.go
```diff
@@ -1,27 +1,79 @@
...
+		})
+
+		Context("when the directories are not exist", func() {
+
+			BeforeEach(func() {
+				dirs = []string{"dir/not/exist/1", "dir/not/exist/2"}
+			})
+
+			It("should return no error", func() {
+
```
steinliber, line 48: This test is empty, If you don't need it, don't put it in the code

internal/pkg/plugin/gitlabcedocker/delete_test.go
```diff
@@ -1,27 +1,79 @@
...
+			})
+		})
+
+		Context("when the directories are exist", func() {
```
steinliber, line 52: for `Context("when the directories are exist"`, you can use `When("the directories are exist")`

internal/pkg/plugin/gitlabcedocker/delete_test.go
aFlyBird0: This is not empty. This test will execute these code:

```go
BeforeEach(func() {
    dirs = []string{"dir/not/exist/1", "dir/not/exist/2"}
})

JustBeforeEach(func() {
	// Remove the directories
	errs = RemoveDirs(dirs)
	for _, e := range errs {
		Expect(e).ToNot(HaveOccurred())
	}
})

It("should return no error", func() {

})

AfterEach(func() {
	// check directories are removed
	for _, dir := range dirs {
		_, err := os.Stat(dir)
		Expect(os.IsNotExist(err)).To(BeTrue())
	}
})

```

I use `BeforeEach` to assign dirs and create dirs, use `justBeforeEach` to execute `RemoveDirs`(the function needs to be tested), and use `AfterEach`  to check if all dirs are deleted.

So, the code in `It` turns out to be empty, it looks a little strange. Do you have any good idea? @steinliber

internal/pkg/plugin/gitlabcedocker/delete_test.go
aFlyBird0: Thank you!

IronCore864 (approved):

algobot76 (approved):
lgtm

Issue #795, opened 2022-07-01
:four_leaf_clover: `Proposal(Need Discussion)`:  Optimize example config file location


aFlyBird0:
## What Would You Like to Add? Why Is This Needed?

Since pr #794 , we could use the command line to show config. But it also brings problems: the config file is duplicated.

<img width="295" alt="image" src="https://user-images.githubusercontent.com/36830265/176816453-2715f94b-ff34-4f4c-a85d-c5bf12769b39.png">

## Design

We have two different solutions now:

### 1. Detele `/examples` and print configs only by `dtm show`.

Advantages:

1. It is easy and clear.

Disadvantages.

1. Config file nesting levels will be  too deep

### 2. update `Makefile`

We duplicated the config files because `go embed` can not embed files from parent directories.

So, why don't we add a `cp` command in `Makefile` to move config files to `/internal/pkg/show/config` ?

Advantages:

1. It is easy for users to reach config files.

Disadvantages.

1. This may be confusing for developers, and we must make it clear that the Makefile is used here to move the configuration file during `make build`.

## Anything else

Please join us to discuss! 🙌

daniel-hutao:
Further, I want all yaml files to be in `examples` directory, including the yaml for each plugin, and make sure to keep only one copy. And the files under `/internal/pkg/show/config` or `/internal/pkg/show/config/plugins` are copied with `make`.

daniel-hutao:
Also, I want to add a `template` directory under the /internal/pkg/show/config/ directory to save the `quickstart.yaml` and `gitops.yaml` and some template yaml files in the future.

@IronCore864 @hxcGit Please join us to discuss~

IronCore864:
I think both solutions are not acceptable, either for first-time users, or for contributors:

1. Only showing config by `dtm show`: new users won't find examples/samples easily. Thinking from a first-time user's perspective, putting a sample config under /examples folder is mandatory. Users would be confused if there is no sample config in an obvious place in the repo.
2. Makefile: it's already complicated; moving more logic to it would only make contributor's life less happy. This solves the duplication but this brings more problems.

Based on the fact that we need "/examples/*.yaml" for new users, and we should only have a single source of yaml to remove duplication, the only logical proposal would be to remove quickstart.yaml from `dtm show`. In this case:

- single source of truth of quickstart.yaml
- if users follow our quickstart guide, they won't use `dtm show` to generate the config for quickstart; it is downloaded by a script.

[...]

Issue #597, opened 2022-05-25
:four_leaf_clover: `Proposal`: More Friendly Quickstart Config Template Output


daniel-hutao:
## What Would You Like to Add? Why is This Needed?

Quickstart, the simpler and easier to get started, the better. So why not have `dtm show config` output a `quickstart.yaml` by default so that users can get started with as little configuration as possible?

## Design

`dtm show config --template=quickstart` -> print `quickstart.yaml`

Let `--template` flag default value == quickstart, so `dtm show config` will also print `quickstart.yaml`

IronCore864:
Approved. Easier for new users to get started without having to copy-paste from the docs.

IronCore864:
Thanks @daniel-hutao and @prodanlabs for the discussion in our group and for raising this issue.

iyear:
ok, I will work for it.

iyear:
What is the `template` flag used for? There is already a `plugin` flag. @daniel-hutao

daniel-hutao:
@iyear the `plugin` flag is used to print config with one specified plugin. The `template` flag is used to print some `combined configs` like `quickstart`/`gitops`/`chatops`/...

iyear:
@daniel-hutao The template files are currently in the `examples` directory, but `show/config/embed.go` cannot read the files in the parent directory.

So the directory planning still needs to consider.

daniel-hutao:
@iyear I'm sorry, This proposal depends on #655, so there are some obstacles when you implement it before. But after #655 merged, I think this proposal is simple enough, let's leave it to NEW

Thor-wl:
Kindly to ask if there is somebody working for this issue? If not, I'm glad to undertake it.
/cc @daniel-hutao

daniel-hutao:
@Thor-wl You got it!!!
Enjoy it!

PR #847, opened 2022-07-13
fix: show non-existent plugin config

### LINKED ISSUE #841: :bomb: `Bug`: command `dtm show config -p harbor` print the wrong config sample
### What Happened?

- `./dtm show config -p harbor`

```sh
---
# core config
varFile: "" # If not empty, use the specified external variables config file
toolFile: "" # If not empty, use the specified external tools config file
state: # state config, backend can be local or s3
  backend: local
  options:
    stateFile: devstream.state

---
# plugins config
tools:
  - name: github-repo-scaffolding-golang
    instanceID: default
    options:
      owner: YOUR_GITHUB_USERNAME_CASE_SENSITIVE
      repo: go-webapp-devstream-demo
      branch: main
      image_repo: YOUR_DOCKER_USERNAME/go-webapp-devstream-demo
  - name: githubactions-golang
    instanceID: default
    dependsOn: ["github-repo-scaffolding-golang.default"]
    options:
      owner: YOUR_GITHUB_USERNAME_CASE_SENSITIVE
      repo: go-webapp-devstream-demo
      language:
        name: go
        version: "1.18"
      branch: main
      build:
        enable: True
      test:
        enable: True
        coverage:
          enable: True
      docker:
        enable: False

```

### How to Reproduce?

./dtm show config -p harbor

### Anything else

_No response_

### DevStream Version

latest

iyear:
## Pre-Checklist

## Description
I remove the default value of `template`. We can discuss better solution.
## Related Issues
close #841 

image

### DIFF
--- a/internal/pkg/show/config/config.go
+++ b/internal/pkg/show/config/config.go
@@ -6,15 +6,21 @@ import (
 	"github.com/spf13/viper"
 )
 
+var templates = map[string]string{
+	"quickstart": QuickStart,
+	"gitops":     GitOps,
+}
+
 //go:generate go run gen_embed_var.go
 func Show() error {
+	// at first, check is template arg is set
 	template := viper.GetString("template")
-	if template == "quickstart" {
-		fmt.Println(QuickStart)
-		return nil
-	} else if template == "gitops" {
-		fmt.Println(GitOps)
-		return nil
+	if template != "" {
+		if tmpl, ok := templates[template]; ok {
+			fmt.Println(tmpl)
+			return nil
+		}
+		return fmt.Errorf("illegal template name : < %s >", template)
 	}
 
 	plugin := viper.GetString("plugin")
--- a/cmd/devstream/show.go
+++ b/cmd/devstream/show.go
@@ -42,14 +42,14 @@ Examples:
 	Run: showStatusCMDFunc,
 }
 
-func showConfigCMDFunc(cmd *cobra.Command, args []string) {
+func showConfigCMDFunc(_ *cobra.Command, _ []string) {
 	log.Debug("Show configuration information.")
 	if err := config.Show(); err != nil {
 		log.Fatal(err)
 	}
 }
 
-func showStatusCMDFunc(cmd *cobra.Command, args []string) {
+func showStatusCMDFunc(_ *cobra.Command, _ []string) {
 	log.Debug("Show status information.")
 	if err := status.Show(configFile); err != nil {
 		log.Fatal(err)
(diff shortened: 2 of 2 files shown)


daniel-hutao:
> In #795 , @IronCore864 thinks that templates are less suitable to be exported by command, but are suitable to be placed under `/examples`. The user gets the yaml file as a download.
> 
> Now two concurrent `quickstart.yaml` and `gitops.yaml` are preventing maintainability. If a solution can be discussed, I can fix it in the meantime.
> 
> @daniel-hutao @aFlyBird0

@iyear I've left a comment under #795. Let's have more discussions there.

daniel-hutao:
For quickstart printing logic, my original design was "when `-p` does not exist, `-t` defaults to equal `quickstart` when `-t` is not set; when `-p` exists, `-t` does not take effect". 

We can follow this idea to fix it first. As for whether to delete the logic of `quickstart` configuration printing, we can wait until the end of the #795 discussion and implement it in a separate pr. @iyear

daniel-hutao (changes requested):
see https://github.com/devstream-io/devstream/pull/847#issuecomment-1185273569

iyear:
@daniel-hutao https://github.com/devstream-io/devstream/blob/2c8ba2d9e1cae2b0b1a1569a7d723dfb4b894173/internal/pkg/show/config/config.go#L22-L24

Then I have to remove this code.
Only one of the `-p` and `-t` args can be allowed to carry a default value.

daniel-hutao:
I looked at the code and we need to prioritize support for `default config` printing, i.e. `dtm show config` can only print [internal/pkg/show/config/default.yaml](https://github.com/devstream-io/devstream/blob/d6875bbf7186cda670334f72f521ca314db29e38/internal/pkg/show/config/default.yaml). so we had to drop the default printing logic for `quickstart`. So, let's just keep the `-t quickstart` support.

@iyear Please also check if the documentation needs to be updated.

Releated issue: #597

internal/pkg/show/config/config.go
```diff
@@ -6,14 +6,16 @@ import (
...
 //go:generate go run gen_embed_var.go
 func Show() error {
-	template := viper.GetString("template")
-	if template == "quickstart" {
-		fmt.Println(QuickStart)
-		return nil
-	} else if template == "gitops" {
-		fmt.Println(GitOps)
+	// at first, check is template parameter is set
+	if tmpl, ok := templates[viper.GetString("template")]; ok {
```
daniel-hutao, line 17: What if a user type: `dtm show config -t abcd`?

It is possible that some exception input checks are needed here to inform the user what parameters are supported when the user enters an incorrect parameter.

daniel-hutao (changes requested):

iyear:
@daniel-hutao Is it possible to review and then merge in? I will start working on #795, otherwise it will cause a conflict.

PR #904, opened 2022-07-25
fix: plugin installed by helm status inconsistent

### LINKED ISSUE #896: :bomb: `Bug`: `Update` interface always be called if a plugin using `util.helm`
### What Happened?

When I was testing the Jenkins family of plugins, I found that every time I executed the `apply` command `dtm` would indicate that the resource had drifted, and then execute the Update operation.

image

The line starting with the `-` sign is the one returned by the `Read` interface.
The line starting with `+` is the original resource in the state file, which is updated by the `Create` or `Update` interface.

To rule out a problem with the jenkins plugin, let's look at the code of the argocd plugin.

image

image

image

You can see that `Read` and `Update` call different resource fetch interfaces. This causes the state file to store "static state" after each call to `Update`, while the latest resource fetched by the `Read` interface call is "all state". This will cause the resource to drift and trigger `Update` each time `apply` is called.

Most of the plugins that use helm to install tools have this problem.

### How to Reproduce?

Execute apply multiple times on a plugin that uses helm to install tools (e.g. argocd).

### Anything else

_No response_

### DevStream Version

latest

steinliber:
Signed-off-by: Meng JiaFeng <jiafeng.meng@merico.dev>

## Pre-Checklist

## Description

Fix inconsistent status error for plugin installed by the helm.

## Related Issues

close #896


internal/pkg/plugininstaller/helm/state.go
```diff
@@ -1,153 +1,48 @@
...
-		helmNameAn, exist := ssAn[helm.GetAnnotationName()]
-		if !exist || helmNameAn != opts.GetReleaseName() {
-			log.Infof("Found unknown daemonSet: %s.", ds.GetName())
-		}
-
-		ready := kubeClient.IsDaemonsetReady(&ds)
-		state.Workflows.AddDaemonset(dsName, ready)
-		log.Debugf("The daemonset %s is %t.", ds.GetName(), ready)
+	for _, ds := range allResource.DaemonSet {
+		state.Workflows.AddStatefulset(ds.Name, ds.Ready)
```
aFlyBird0, line 42: a small mistake.

`AddAddDaemonset`

aFlyBird0 (changes requested):

aFlyBird0 (approved):

iyear (approved):

daniel-hutao (approved):
lgtm

PR #923, opened 2022-07-28
feat: makefile-enhancement

jxs1211:
Signed-off-by: xian-jie.shen <327411586@qq.com>

## Pre-Checklist

## Description
There is no go file in the deleted directory, so remove useless code

## Related Issues
N/A

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ DTM_ROOT=github.com/devstream-io/devstream
 GO_LDFLAGS += -X '$(DTM_ROOT)/internal/pkg/version.Version=$(VERSION)' \
 		-X '$(DTM_ROOT)/cmd/devstream/list.PluginsName=$(PLUGINS)'
 
-FIND := find . -path './cmd/**/*.go' -o -path './test/**/*.go' -o -path './pkg/**/*.go' -o -path './internal/**/*.go'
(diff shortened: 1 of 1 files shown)


aFlyBird0:
Thank you for reading the source code so carefully! @jxs1211 

Although there is no go file in the `test` folder now, there may be one later (integration tests, testing functions in different packages).

@daniel-hutao , what is your opinion?

daniel-hutao:
> Thank you for reading the source code so carefully! @jxs1211
> 
> Although there is no go file in the `test` folder now, there may be one later (integration tests, testing functions in different packages).
> 
> @daniel-hutao , what is your opinion?

@aFlyBird0 You are right.

@jxs1211 Thank you for your meticulous work. But we don't need to make this change, because it is possible that the test directory has go files, it has go files there before, and it will probably has go files there later, it just happens no go files there at the moment. Let's leave it as it is.

jxs1211:
> > Thank you for reading the source code so carefully! @jxs1211
> > Although there is no go file in the `test` folder now, there may be one later (integration tests, testing functions in different packages).
> > @daniel-hutao , what is your opinion?
> 
> @aFlyBird0 You are right.
> 
> @jxs1211 Thank you for your meticulous work. But we don't need to make this change, because it is possible that the test directory has go files, it has go files there before, and it will probably has go files there later, it just happens no go files there at the moment. Let's leave it as it is.

got it

PR #930, opened 2022-08-01
feat: copy examples to templates dir

### LINKED ISSUE #795: :four_leaf_clover: `Proposal(Need Discussion)`:  Optimize example config file location

iyear:
Signed-off-by: iyear <ljyngup@gmail.com>

## Pre-Checklist

## Description
Copy the templates with `go generate`

## Related Issues
close #795 

image


daniel-hutao (approved):
@iyear Good job!

IronCore864 (approved):

PR #921, opened 2022-07-28
feat: new plugin jenkins-github-integ

aFlyBird0:
Signed-off-by: Bird <aflybird0@gmail.com>

## Pre-Checklist

Note: please complete **_ALL_** items in the following checklist.

- [x] I have read through the [CONTRIBUTING.md](https://github.com/devstream-io/devstream/blob/main/CONTRIBUTING.md) documentation.
- [x] My code has the necessary comments and documentation (if needed).
- [x] I have added relevant tests

## Description
image

## Related Issues
#920

### config
```yaml
---
# core config
varFile: "" # If not empty, use the specified external variables config file
toolFile: "" # If not empty, use the specified external tools config file
state: # state config, backend can be local or s3
  backend: local
  options:
    stateFile: devstream-jenkins-all.state

---

tools:
  # name of the tool
  - name: jenkins
    # id of the tool instance
    instanceID: default
    # format: name.instanceID; If specified, dtm will make sure the dependency is applied first before handling this tool.
    dependsOn: [ ]
    # options for the plugin
    options:
      # if true, the plugin will use hostpath to create a pv named `jenkins-pv`
      # and you should create the volumes directory manually, see plugin doc for details.
      test_env: true
      # need to create the namespace or not, default: false
      create_namespace: true
      # Helm repo information
      repo:
        # name of the Helm repo
        name: jenkins
        # url of the Helm repo
        url: https://charts.jenkins.io
      # Helm chart information
      chart:
        # name of the chart
        chart_name: jenkins/jenkins
        # release name of the chart
        release_name: dev
        # k8s namespace where jenkins will be installed
        namespace: jenkins
        # whether to wait for the release to be deployed or not
        wait: true
        # the time to wait for any individual Kubernetes operation (like Jobs for hooks). This defaults to 5m0s
        timeout: 10m
        # whether to perform a CRD upgrade during installation
        upgradeCRDs: true
        # custom configuration. You can refer to [Jenkins values.yaml](https://github.com/jenkinsci/helm-charts/blob/main/charts/jenkins/values.yaml)
        values_yaml: |
          persistence:
            # for prod env: the existent storageClass, please change it
            # for test env: just ignore it, but don't remove it
            storageClass: jenkins-pv
          serviceAccount:
            create: false
            name: jenkins
          controller:
            serviceType: NodePort
            nodePort: 32000
            additionalPlugins:
              # install "GitHub Pull Request Builder" plugin, see https://plugins.jenkins.io/ghprb/ for more details
              - ghprb
[...]

PR #939, opened 2022-08-03
fix: fail to create github webhook automatically

aFlyBird0:

## Description
I created a new build in #921 to automatically create a Jenkins Job triggered by GitHub.

This pr is here to fix it.

Although the code looks weird, in the short term, it's the easiest and quickest way to fix it.

After a lot of troubleshooting, here's why I couldn't automatically generate the webhook

The first step is to create a ConfigMap to generate JCasC, which has a setting `manageWebhooks: true` to automatically create a webhook for the GitHub repo when you create a job, and the second step is to create a Jenkins job.

The problem is that after creating the ConfigMap in step 1, you need to poll the ConfigMap via sidecar to convert it to JCasC, so the `manageWebhooks` configuration takes effect after the job is created.

We just need to make sure that we create the job after the configuration takes effect. The current strategy is to sleep for 3 seconds.

Of course, the best strategy is to poll the latest JCasC and create the job as soon as the configuration takes effect, but considering that we need to spend time testing the project near the release, and that Jenkins installation itself is a very time-consuming task, it is acceptable to wait three seconds longer than that.

## Related Issues
#921

### DIFF
--- a/internal/pkg/plugin/jenkinsgithub/jcasc.go
+++ b/internal/pkg/plugin/jenkinsgithub/jcasc.go
@@ -6,6 +6,7 @@ import (
 	_ "embed"
 	"fmt"
 	"text/template"
+	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1 "k8s.io/client-go/applyconfigurations/core/v1"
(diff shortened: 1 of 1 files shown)


iyear (approved):

aFlyBird0:
@steinliber PTAL 🙌

PR #1021, opened 2022-08-22
Feat: Remove "test env" Logic from the `jenkins` Plugin

### LINKED ISSUE #1020: :four_leaf_clover: `Proposal`: Remove the `test env` Logic from `jenkins` Plugin
## What Would You Like to Add? Why Is This Needed?

1. Configuring `StorageClass` is easy;
   1. Most users will use `minikube` when testing k8s, after all, this is what recommended on the k8s document;
   2. The `minikube` deployment of a k8s cluster includes a `default StorageClass`, and users do not need to think about how to configure the StorageClass;
2. Use `StorageClass` will make our code cleaner, more consistent, and easier to understand;
   1. With the `helm chart`, we can easily configure `StorageClass` to persist data;
3. With `test_env` we want to help users do more configuration work in the test environment, and then there is still no way around reminding users to run commands like `chmod` through the documentation. So the added complexity still doesn't get 100% of the work done, and is probably more of a burden to users.

## Design

Use `StorageClass` in any environment.

## Anything else

I'd like to do it.

daniel-hutao:

## Pre-Checklist

## Description

1. Remove "test env" logic from the `jenkins` plugin. See #1020 for more info.
2. Remove preCreate/postDelete logic. `ServiceAccount` could be created by helm.
3. Remove "show url" logic too. `NodePort` is just a test scenario, I will add `NodePort` and `Ingress` scenarios to the documentation to make the logic related to "service expose" more general in the next pr.

I'll do more tests against this plugin and add more documents in the next pr.

@IronCore864 @aFlyBird0 Please take a review at this pr. thx.

## Related Issues

Close #1020

No test_env option from now on.

I haven't test this change locally for the bad network. But I will continue to think about how to optimize this plugin and will do more tests before the next pr.


aFlyBird0 (approved):
LGTM. 👍

internal/pkg/show/config/plugins/jenkins.yaml
```diff
@@ -22,8 +19,6 @@ tools:
       chart:
         # name of the chart
         chart_name: jenkins/jenkins
-        # release name of the chart
-        release_name: dev
```
aFlyBird0, line 26: Is it on purpose or by mistake? @daniel-hutao

daniel-hutao: purpose

IronCore864 (approved):

Issue #1084, opened 2022-09-14
:four_leaf_clover: `Proposal`: Use DevStream in an Air-gapped Environment


daniel-hutao:
## What Would You Like to Add? Why Is This Needed?

See #1082 for more background info.

## Design

- [x] Add a script for `dtm` and `dtm plugins` downloading @aFlyBird0 
- [x] Add a script for `helm charts` downloading @aFlyBird0 
- [ ] `dtm plugins` customize image repository supproting @daniel-hutao 
- [x] `dtm` local charts installation supporting @daniel-hutao 
- [ ] `Jenkins`: add a script for `Jenkins plugins` downloading @steinliber 
- [ ] Add some documents for the `dtm` air-gapped using @daniel-hutao 
- [ ] List all images needed by `jenkins` and `harbor` @daniel-hutao 

## Anything else

Go Go Go ^-^

PR #1104, opened 2022-09-18
feat: use `dtm init --download-only` to download specify plugins from flags

aFlyBird0:

## Pre-Checklist

## Description
1. use `dtm init --download-only` to download specify plugins from flags.
2. use `--plugins` to specify plugins to download.
3. use `--all` to download all plugins.
4. use `--os` and `--arch` to specify os and arch, the default is os and arch of the machine which executes dtm.

## Related Issues
The first task of #1084.

(We can use quick-start script to download `dtm` and use `dtm init --all --os=DestinyOS --arch=DestinyArch -d=PLUGIN_DIR` to download plugins.

Comment out this code and test:

image

image

image

image

### DIFF
--- a/cmd/devstream/init.go
+++ b/cmd/devstream/init.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"runtime"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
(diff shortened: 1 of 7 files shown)


daniel-hutao:
@aFlyBird0 Hi bird, I have a small question here. Why do we need `--all` flag here?

aFlyBird0:
> @aFlyBird0 Hi bird, I have a small question here. Why do we need `--all` flag here?

We need to add a flag to download all plugins so that users could use `dtm` offline. If `--all` flag is set, config file can be empty.

If not, users must create a config file and execute `dtm init`.

daniel-hutao:
> > @aFlyBird0 Hi bird, I have a small question here. Why do we need `--all` flag here?
> 
> We need to add a flag to download all plugins so that users could use `dtm` offline. If `--all` flag is set, config file can be empty.
> 
> If not, users must create a config file and execute `dtm init`.

That is, if a user needs to use 3 plugins only, then we suggest he/she download all plugins.

Maybe there's nobody who needs to use all plugins. Do most users need to download just 1-5? plugins, but dtm has 28 plugins(maybe more than 100 plugins next year).

So my question is: what if a user needs to download 5/100 plugins, and what should he/she do?

daniel-hutao:
How about we use `dtm init --download-only --plugins="jenkins, harbor, gitlab"` to download plugins? `--all` flag equals false by default.

aFlyBird0:
> How about we use `dtm init --download-only --plugins="jenkins, harbor, gitlab"` to download plugins? `--all` flag equals false by default.

Cool, I agree with you! 🚀

cmd/devstream/init.go
```diff
@@ -20,35 +24,119 @@ var initCMD = &cobra.Command{
...
+		log.Fatalf("Dev version plugins can't be downloaded from the remote plugin repo; please run `make build-plugin.PLUGIN_NAME` to build them locally.")
+	}
+
+	var (
+		realPluginDir string
+		tools         []configmanager.Tool
+		err           error
+	)
+
+	switch downloadOnly {
```
daniel-hutao, line 46:
```go
	switch downloadOnly {
 	// download plugins according to the config file
 	case false:
 		tools, realPluginDir, err = GetPluginsAndPluginDirFromConfig()
 	// download plugins from flags
 	case true:
 		tools, realPluginDir, err = GetPluginsAndPluginDirFromFlags()
 	}
```

vs

```go
if downloadOnly {
    tools, realPluginDir, err = GetPluginsAndPluginDirFromFlags()
} else {
    tools, realPluginDir, err = GetPluginsAndPluginDirFromConfig()
}
```

I think the "if" version is more clear here.

cmd/devstream/init.go
```diff
@@ -20,35 +24,119 @@ var initCMD = &cobra.Command{
...
 }
 
 func init() {
+	// flags for init from config file
 	initCMD.Flags().StringVarP(&configFile, configFlagName, "f", "config.yaml", "config file")
 	initCMD.Flags().StringVarP(&pluginDir, pluginDirFlagName, "d", "", "plugins directory")
 
+	// downloading specific plugins from flags
+	initCMD.Flags().BoolVar(&downloadOnly, "download-only", false, "download plugins only")
+	initCMD.Flags().StringSliceVarP(&pluginsToDownload, "plugins", "p", []string{}, "plugins to download")
```
daniel-hutao, line 135: "the plugins to be downloaded"?

daniel-hutao (changes requested):

cmd/devstream/init.go
aFlyBird0: That sounds great.

daniel-hutao (approved):

Issue #941, opened 2022-08-04
:open_book: `Docs`: The Full Configuration Examples Need to be Added to the Plugin Documentations


daniel-hutao:
### What should be changed?

A long time ago, we had only one configuration file, in a format that probably looked like below:

```yaml
tools:
  foo: bar
```

So at that time, the configuration examples given in our plugin documentations are the same as above.

But now the configuration file has been split into 3 parts: `core config`, `var config` and `tools config`. So when a user opens one of our plugin documents and sees the example like below, he won't know how to `apply` it.

![image](https://user-images.githubusercontent.com/18692628/182821641-8190e9fc-6be4-4eb4-9e1b-2f90c86a62d7.png)

I think all the plugin documentations about the configuration examples should be changed to like below:

```yaml
---
# core config
varFile: "" # If not empty, use the specified external variables config file
toolFile: "" # If not empty, use the specified external tools config file
state: # state config, backend can be local or s3
  backend: local
  options:
    stateFile: devstream.state

---
# variables config
foo: bar

---
# plugins config
tools:
  - name: foo
    instanceID: default
    options:
      fkk: [[ foo ]]
```

### Please read the documents below.

- [X] I've read the document [Creating a Documentation for DevStream](https://docs.devstream.io/en/latest/development/mkdocs/) and [Document Translation(Chinese only)](https://docs.devstream.io/en/latest/development/translation.zh/). (If you want to submit a document type contribution.)

daniel-hutao:
@IronCore864 @steinliber @aFlyBird0 @iyear Please leave some comments below.

iyear:
But it's too redundant, and if our config file changes in the future, every plugin will have to be changed. 

It can be explained on the `Plugin` overview page.

daniel-hutao:
> If we want to change core config or var config definition, we have to change all docs. I think maybe we can define `core`, `var`, and `tool` config in one docs, And plugin docs only describe it's config items

This is a good idea, we add a document to introduce the usage of configuration files, and then in each plugin documentation to declare the current configuration is only "tools config", and tell the users to refer to the documentation we just added for how to write the full configuration files.

daniel-hutao:
I'll do it. cc @IronCore864

IronCore864:
Let me split this ticket up into a few smaller ones and we can share the workload.

IronCore864:
> I'll do it. cc @IronCore864

It seems there isn't much progress on this issue for a while. I'll try to finish it today.

PR #1153, opened 2022-10-09
Docs update plugins

IronCore864:
## Description

Update plugins documentation:

- title case
- explicitly mention that the config example is only the tool file, and add links to var file, main config docs

## Related Issues

Closes https://github.com/devstream-io/devstream/issues/941.

### DIFF
--- a/internal/pkg/show/config/plugins/jenkins-pipeline.yaml
+++ b/internal/pkg/show/config/plugins/jenkins-pipeline.yaml
@@ -9,43 +9,43 @@ tools:
   options:
     jenkins:
       # url is used to config jenkins url
-      url: http://jenkins.example.com:8080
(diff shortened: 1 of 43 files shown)


internal/pkg/show/config/plugins/jenkins-pipeline.yaml
```diff
@@ -9,43 +9,43 @@ tools:
   options:
     jenkins:
       # url is used to config jenkins url
-      url: http://jenkins.example.com:8080
```
daniel-hutao, line 12: If we change all the example configurations to placeholders, the complexity of using them will be much higher. The same configuration file would require more modifications.

daniel-hutao (changes requested):

Issue #1491, opened 2023-04-24
:four_leaf_clover: `Proposal`: We Are Back


daniel-hutao:
After more than two months of hibernation, today we decided to awaken DevStream for its second life. I will redevelop DevStream on a new branch: new-dtm. Perhaps in a month, the brand-new DevStream will meet everyone. At that time, I will merge the code from the new-dtm branch into main and reactivate the DevStream community.

jolarsso:
This seems like a really cool tool that I (at least partly) would like to include in a fututure DevOps workflow. But looking at the repo, last commit was two years ago? 

What is the current status of the project?

PR #1492, opened 2023-04-24
Feat: This Is a Special Version.

daniel-hutao:
This is a special version. The current version has removed all old code and now only provides the dtm patch command.


daniel-hutao:
#1491

